PR #1, opened 2025-10-10
[IT-4653] Initial PR of all code from susheel's MCP repo into sage main

thomasyu888:
## Problem
We want the code in Sage/synapse-mcp repo, but we also want to review this code before it gets pushed into `main` branch

## Solution

I created a new blank repo with a simple readme, followed these to import susheel's code, then created a PR.

```
git clone git@github.com:Sage-Bionetworks/synapse-mcp.git
cd synapse-mcp
git remote add upstream https://github.com/susheel/synapse-mcp.git
git fetch upstream
git checkout -b import-upstream
git merge upstream/main --allow-unrelated-histories
git checkout upstream/main -- README.md
git add README.md
git commit -m "Initial import of code from susheel/synapse-mcp"
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements-dev.txt
+
+      - name: Run pytest
+        env:
+          SYNAPSE_PAT: dummy-token
+          TMPDIR: ${{ runner.temp }}
+        run: pytest
+
+      - name: Docker build
+        run: docker build --tag synapse-mcp .
--- a/COMPLIANCE.md
+++ b/COMPLIANCE.md
@@ -1,0 +1,30 @@
+## Data Privacy and Terms of Service Compliance
+
+### Understanding the Risk
+
+This MCP server enables AI assistants to query Synapse biomedical datasets using your personal access token. However, the [Synapse Terms of Service](https://www.synapse.org/TrustCenter:TermsOfService) explicitly prohibit redistribution of Synapse data.
+
+When you use this server with external, cloud-based AI models:
+1. Your queries and Synapse's responses are sent to the AI provider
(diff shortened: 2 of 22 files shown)


src/synapse_mcp/entities/base.py
```diff
@@ -0,0 +1,85 @@
...
+        Args:
+            entity_id: The Synapse ID of the entity
+
+        Returns:
+            The entity as a dictionary
+        """
+        # IMPORTANT: Always keep downloadFile=False to avoid downloading file content
+        # when fetching entity metadata. This prevents unnecessary downloads and
+        # keeps the operation lightweight for metadata-only operations.
+        entity = self.synapse_client.get(entity_id, downloadFile=False)
```
BryanFauble: There is an opportunity to swap over to https://python-docs.synapse.org/en/stable/reference/experimental/sync/factory_operations/#synapseclient.operations.get - It would have some downstream affects that also need to be addressed like swapping the `format_entity` over to use the new `snake_case` attributes. That has the consequence that the functions to list children of the project/folder also need to be swapped as well since they all eventually call to `format_entity`.

src/synapse_mcp/oauth/client_registry.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def load_all(self) -> list[ClientRegistration]:
+        with self._lock:
+            if not self._path.exists():
+                return []
+            try:
+                data = json.loads(self._path.read_text())
+            except json.JSONDecodeError as exc:  # pragma: no cover - defensive
+                logger.warning("Failed to parse client registry file %s: %s", self._path, exc)
+                return []
```
BryanFauble: In cases across the MCP server where the exception is being included in the logger string, we can also do something like:

```suggestion
            except json.JSONDecodeError:  # pragma: no cover - defensive
                logger.warning(
                    "Failed to parse client registry file %s.", self._path, exc_info=True)
                return []
```

This will capture the stack trace for the exception automatically. Giving us a lot more context to debug issues as they do pop up.

src/synapse_mcp/session_storage/redis_backend.py
```diff
@@ -0,0 +1,233 @@
...
+                    self.redis_url,
+                    decode_responses=True,
+                    socket_connect_timeout=5,
+                    socket_timeout=5,
+                    health_check_interval=30,
+                )
+                await self._redis.ping()
+                logger.info("Redis connection established")
+            except Exception as exc:  # pragma: no cover - connection failure
+                logger.error("Failed to connect to Redis: %s", exc)
```
BryanFauble: In addition to setting `exc_info=True` in the log message, using `.exception` will do that automatically and emit the log at Error level.

```suggestion
                logger.exception("Failed to connect to Redis")
```

PR #9, opened 2025-11-25
[DPE-1500] Refactor JWT verification to use instance algorithm

BryanFauble:
# **Problem:**

- When setting up the deployed MCP server in AWS ECS with Oauth enabled it was erroring out auth with `synapse_mcp.oauth - ERROR - Error in async Synapse JWT verification: 'PyJWK' object has no attribute 'algorithm_name'`

# **Solution:**

- Use the static RS256 as that is what the clients are registered with. If need be we can revist this later.

# **Testing:**

- Will build an image and deploy to ECS after mge


thomasyu888 (approved):
🔥 LGTM!

PR #24, opened 2026-04-09
[DPE-1623] Migrate legacy entity tools to service-layer pattern

BryanFauble:
## Summary

Replaces 5 legacy entity tools that used direct `synapseclient.Synapse` method calls with typed SDK model classes via the new service layer. Deletes the entire `entities/` directory. Adds 9 new entity-related tools. PR 2 of 7 in [DPE-1621](https://sagebionetworks.jira.com/browse/DPE-1621).

**Jira:** [DPE-1623](https://sagebionetworks.jira.com/browse/DPE-1623)

### Migrated tools

| Tool | Before | After |
|------|--------|-------|
| `get_entity` | `client.get(id, downloadFile=False)` | `operations.get(id, FileOptions(download_file=False))` |
| `get_entity_annotations` | `client.get_annotations(id)` | `operations.get(id, ...)` then `.annotations` |
| `get_entity_children` | `client.getChildren(id)` | `Folder(id).walk(recursive=False)` |
| `get_entity_provenance` | `client.getProvenance(id)` | `Activity.get(parent_id=id)` |
| `search_synapse` | inline `client.restPOST` | `SearchService` with `@error_boundary` |

### New tools

`get_entity_acl`, `get_entity_permissions`, `list_entity_acl`, `get_entity_schema`, `get_entity_schema_derived_keys`, `get_entity_schema_validation_statistics`, `get_entity_schema_invalid_validations`, `sync_container`, `get_link`, `get_activity`

### Deleted

- `src/synapse_mcp/entities/` (7 files, ~290 lines)
- `get_entity_operations()` from `context_helpers.py`
- `format_synapse_entity()`, `format_annotations()` from `utils.py`

## Architecture

```mermaid
flowchart LR
    subgraph Before
        T1[tools.py] --> EO[get_entity_operations]
        EO --> E1[entities/base.py]
        EO --> E2[entities/project.py]
        E1 --> SC1["client.get()"]
        E2 --> SC2["client.getChildren()"]
    end

    subgraph After
        T2[tools.py] --> ES[EntityService]
        T2 --> AS[ActivityService]
        T2 --> SS[SearchService]
        ES --> OG["operations.get(FileOptions)"]
        ES --> FW["Folder.walk()"]
        AS --> AG["Activity.get()"]
        SS --> RP["client.restPOST('/search')"]
    end
```

### File download safety

```mermaid
flowchart TD
    A[SDK method call] --> B{Can it download files?}
    B -->|Yes| C{Has download_file param?}
    B -->|No| D[Safe - metadata only]
    C -->|Yes| E["Always pass download_file=False"]
    C -->|No| F[Excluded from scope]
    
    E --> G["operations.get(FileOptions(download_file=False))"]
    E --> H["sync_from_synapse(download_file=False)"]
    E --> I["operations.get(LinkOptions, FileOptions)"]
    F --> J["query() - excluded"]
    F --> K["RecordSet.get_detailed_validation_results() - excluded"]
```

## Test plan

- [x] Entity get/annotations/children tests updated for new SDK patterns
- [x] Activity service tests (provenance by entity, by activity ID)
- [x] Search service tests (payload construction, field rejection fallback)
- [x] Curation task service + manager tests migrated
- [x] All 112 tests pass

[DPE-1623]: https://sagebionetworks.jira.com/browse/DPE-1623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,358 @@
...
+        Returns:
+            Dict with entity_id, principal_id, and a list
+            of access_types (e.g. READ, UPDATE, DELETE).
+        """
+        async with synapse_client(ctx) as client:
+            entity = File(id=entity_id)
+            access_types = await entity.get_acl_async(
+                principal_id=principal_id,
+                synapse_client=client,
+            )
```
BryanFauble: Look into why this is a File class instance when getting the acl, especially when the entity is not a FIle.

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,374 @@
...
+            types. Returns an error dict inside a list if
+            the entity is not a container.
+        """
+        async with synapse_client(ctx) as client:
+            container = Folder(id=entity_id)
+            await container.sync_from_synapse_async(
+                download_file=False,
+                recursive=False,
+                synapse_client=client,
+            )
```
BryanFauble, line 156: When this sync is being performed it is not including all possible child types.

src/synapse_mcp/tools.py
```diff
@@ -1,205 +1,128 @@
...
         return {"error": f"Invalid Synapse ID: {entity_id}"}
-
-    try:
-        entity_ops = await get_entity_operations(ctx)
-        return entity_ops["base"].get_entity_by_id(entity_id)
-    except ConnectionAuthError as exc:
-        return {"error": f"Authentication required: {exc}", "entity_id": entity_id}
-    except Exception as exc:  # pragma: no cover - defensive path
-        return {"error": str(exc), "entity_id": entity_id}
+    return await EntityService().get_entity(ctx, entity_id)
```
BryanFauble: This is a bit awkward to need to create an instance of EntityService, it would make more sense for the methods to be static.

src/synapse_mcp/services/activity_service.py
```diff
@@ -0,0 +1,95 @@
...
+                    "entity_id": entity_id,
+                    "version": version,
+                }
+            result: Dict[str, Any] = {
+                "entity_id": entity_id,
+                "activity": serialize_model(activity),
+            }
+            if version is not None:
+                result["version"] = version
+            return result
```
BryanFauble: This is duplicated with the `get_activity` functionality, we really only need one. We can mentioned provenance and activity in the description.

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,374 @@
...
+    "files",
+    "folders",
+    "tables",
+    "entityviews",
+    "submissionviews",
+    "datasets",
+    "datasetcollections",
+    "materializedviews",
+    "virtualtables",
+)
```
BryanFauble, line 61: Determine if we can remove this and dynamically grab a list of entity types instead of having a hard coded list here https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityType.html

src/synapse_mcp/tools.py
```diff
@@ -211,175 +134,274 @@ async def search_synapse(
...
+
+@mcp.tool(
+    title="Get Activity",
+    description=(
+        "Get a provenance Activity by its own ID, "
+        "or by parent entity ID and optional version."
+    ),
+    annotations=_RO,
+)
+async def get_activity(
```
BryanFauble: This tool can be dropped out in favor of the existing `get_entity_provenance` call instead. The `get_entity_provenance` can have it's description mention that "Activity" is the synapse specific name for this concept.

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,390 @@
...
+            ctx: The FastMCP request context.
+            entity_id: Synapse ID of a Project or Folder.
+
+        Returns:
+            List of child entity dicts covering all entity
+            types. Returns an error dict inside a list if
+            the entity is not a container.
+        """
+        async with synapse_client(ctx) as client:
+            container = Folder(id=entity_id)
```
andrewelamb, line 151: Won't this cause an error if the user inputs a Project ID?

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,390 @@
...
+            ctx: The FastMCP request context.
+            entity_id: Synapse ID of a Folder or Project
+                with a bound JSON schema.
+
+        Returns:
+            Dict with validation statistics (counts of
+            valid, invalid, unknown entities).
+        """
+        async with synapse_client(ctx) as client:
+            container = Folder(id=entity_id)
```
andrewelamb, line 339: Same as above

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,390 @@
...
+            ctx: The FastMCP request context.
+            entity_id: Synapse ID of a Folder or Project
+                with a bound JSON schema.
+
+        Returns:
+            List of dicts describing entities that failed
+            validation.
+        """
+        async with synapse_client(ctx) as client:
+            container = Folder(id=entity_id)
```
andrewelamb, line 368: Same as above

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,390 @@
...
+            results = await container.get_invalid_validation_async(
```
andrewelamb: Since this is an AsyncGenerator should this be awaited?

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,390 @@
...
+
+        Returns:
+            Dict with entity_acl (current entity's ACL
+            entries) and all_entity_acls (includes
+            descendants if recursive).
+        """
+        async with synapse_client(ctx) as client:
+            entity = await _resolve_entity(entity_id, client)
+            acl_result = await entity.list_acl_async(
+                recursive=recursive,
```
andrewelamb, line 260: Should this include `include_container_content=recursive`?

src/synapse_mcp/services/entity_service.py
BryanFauble: No it will not cause an error

src/synapse_mcp/services/entity_service.py
BryanFauble: No

src/synapse_mcp/services/entity_service.py
BryanFauble: No

src/synapse_mcp/services/entity_service.py
BryanFauble: Good catch — that was a bug. `get_invalid_validation_async` is an `AsyncGenerator`, not a coroutine, so awaiting it returned the generator object instead of driving it. Fixed: the call is now `async for item in container.get_invalid_validation_async(...)`. Added a regression test in `tests/test_entity_service.py::TestGetSchemaInvalidValidations` that yields two records and asserts both come back.

src/synapse_mcp/services/entity_service.py
BryanFauble: Yes, you're right. Per the SDK docstring, `recursive=True` requires `include_container_content=True` and the SDK raises `ValueError` otherwise. Fixed: `list_acl` now exposes both `include_container_content` and `target_entity_types` parameters and pre-validates the combo so the caller gets a clean error dict before the SDK is invoked. The MCP tool wrapper `list_entity_acl` was updated to pass them through. Regression tests added in `tests/test_entity_service.py::TestListAcl`.

src/synapse_mcp/tools.py
```diff
@@ -1,205 +1,145 @@
...
-    entity_id: str,
     ctx: Context,
+    entity_id: Optional[str] = None,
     version: Optional[int] = None,
+    activity_id: Optional[str] = None,
 ) -> Dict[str, Any]:
-    """Return activity metadata for a Synapse entity, optionally scoping to a specific version."""
-    if not validate_synapse_id(entity_id):
+    """Return provenance/activity metadata for a Synapse entity."""
+    if entity_id is None and activity_id is None:
```
linglp, line 92: Hmm this duplicated the logic in `ActivityService.get_provenance`

src/synapse_mcp/tools.py
```diff
@@ -211,175 +151,248 @@ async def search_synapse(
...
-        "facets": response.get("facets", []),
-        "query": request_payload,
-    }
+# ---------------------------------------------------------------------------
+# Domain 2: Entity Access Control
+# ---------------------------------------------------------------------------
+
+
+@mcp.tool(
+    title="Get Entity ACL",
```
linglp, line 181: Nit: I am not sure if this is an inconsistency that we want to address. But other mcp tools such as `async def get_entity_annotations` has title `title="Fetch Entity Annotations",` But here we are using `Get` instead of `Fetch`.

tests/test_entity_service.py
```diff
@@ -0,0 +1,390 @@
...
+class TestGetEntity:
+    @patch(f"{TS}.get_synapse_client", new_callable=AsyncMock)
+    @patch(f"{SVC}.operations_get_async", new_callable=AsyncMock)
+    async def test_given_valid_entity_when_fetched_then_returns_serialized_dict(
+        self, mock_ops_get, mock_get_client
+    ):
+        mock_get_client.return_value = MagicMock()
+        mock_ops_get.return_value = FakeEntity()
+
+        result = await EntityService().get_entity(MagicMock(), "syn123")
```
linglp, line 67: Nit: This test passed. But I think we don't need to do `EntityService()` here and it can be confusing because `get_entity` is a static method which means that we could call `EntityService.get_entity(ctx, id)` directly without instantiating

tests/test_activity_service.py
```diff
@@ -0,0 +1,160 @@
...
+    @patch(f"{SVC}.Activity")
+    async def test_given_entity_with_provenance_when_fetched_then_returns_activity(
+        self, mock_activity_cls, mock_get_client
+    ):
+        # GIVEN an entity that has provenance
+        mock_get_client.return_value = MagicMock()
+        mock_activity_cls.get_async = AsyncMock(return_value=FakeActivity())
+
+        # WHEN we get provenance
+        result = await ActivityService().get_provenance(
```
linglp, line 60: Nit: similarly, here we could call `ActivityService.get_provenance` directly without instantiating,

tests/test_entity_service.py
```diff
@@ -0,0 +1,390 @@
...
+class FakeAclResult:
+    entity_acl: Optional[dict] = None
+    all_entity_acls: Optional[list] = None
+
+
+class TestListAcl:
+    @patch(f"{TS}.get_synapse_client", new_callable=AsyncMock)
+    @patch(f"{SVC}.operations_get_async", new_callable=AsyncMock)
+    async def test_given_recursive_without_container_content_then_returns_error(
+        self, mock_ops_get, mock_get_client
```
linglp, line 312: Nit: mock_get_client can be removed?

src/synapse_mcp/tools.py
```diff
@@ -211,175 +151,248 @@ async def search_synapse(
...
+    ),
+    annotations=_RO,
+)
+async def list_entity_acl(
+    entity_id: str,
+    ctx: Context,
+    recursive: bool = False,
+    include_container_content: bool = False,
+    target_entity_types: Optional[List[str]] = None,
+) -> Dict[str, Any]:
```
linglp, line 236: Is this the right return type even when `recursive = True`? I think if `recursive=True`, a list of ACL will be returned?

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,414 @@
...
+            Dict with validation statistics (counts of
+            valid, invalid, unknown entities).
+        """
+        async with synapse_client(ctx) as client:
+            container = Folder(id=entity_id)
+            stats = await container.get_schema_validation_statistics_async(
+                synapse_client=client,
+            )
+            result = serialize_model(stats)
+            if isinstance(result, dict):
```
linglp, line 344: Nit: Do we need to be defensive here? It seems like a dictioary will always be returned.

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,414 @@
...
+            Dict with JSON schema binding info
+            (organization, schema name, version, etc.).
+        """
+        async with synapse_client(ctx) as client:
+            entity = await _resolve_entity(entity_id, client)
+            schema_info = await entity.get_schema_async(
+                synapse_client=client,
+            )
+            result = serialize_model(schema_info)
+            if isinstance(result, dict):
```
linglp, line 292: Nit: Do we need to be defensive here? It seems like it will always return a dictionary.

linglp (approved):
@BryanFauble LGTM! Just some minor comments.

src/synapse_mcp/services/entity_service.py
```diff
@@ -0,0 +1,414 @@
...
+        Returns:
+            List of dicts describing entities that failed
+            validation.
+        """
+        async with synapse_client(ctx) as client:
+            container = Folder(id=entity_id)
+            # ``get_invalid_validation_async`` is an AsyncGenerator,
+            # not a coroutine — drive it with ``async for`` rather
+            # than awaiting it.
+            return [
```
linglp, line 372: Should we return all the results in a list? Or should we use `collect_generator`?

PR #25, opened 2026-04-09
[DPE-1624] Add Wiki service and tools

BryanFauble:
## Summary

Adds 4 wiki tools so LLM agents can read wiki content, navigate page trees, review revision history, and understand page ordering. PR 3 of 7 in [DPE-1621](https://sagebionetworks.jira.com/browse/DPE-1621).

**Jira:** [DPE-1624](https://sagebionetworks.jira.com/browse/DPE-1624)

## New tools

| Tool | Purpose | SDK Method |
|------|---------|-----------|
| `get_wiki_page` | Get markdown content + metadata | `WikiPage(owner_id, id).get()` |
| `get_wiki_headers` | Get page tree (TOC) | `WikiHeader.get(owner_id)` |
| `get_wiki_history` | Get revision history | `WikiHistorySnapshot.get(owner_id, id)` |
| `get_wiki_order_hint` | Get sub-page ordering | `WikiOrderHint(owner_id).get()` |

## Root page auto-discovery

The SDK requires `id` or `title` for `WikiPage.get()` — it can't fetch the root page from just `owner_id`. The service handles this automatically:

```mermaid
flowchart TD
    A["get_wiki_page(owner_id, wiki_id=None)"] --> B{wiki_id provided?}
    B -->|Yes| F["WikiPage(owner_id, id=wiki_id).get()"]
    B -->|No| C["WikiHeader.get(owner_id)"]
    C --> D["Find header with parent_id=None"]
    D --> E{Root found?}
    E -->|Yes| F2["WikiPage(owner_id, id=root.id).get()"]
    E -->|No| G["Return error: no root wiki page"]
```

## Test plan

- [x] get_wiki_page with specific wiki_id
- [x] get_wiki_page auto-discovers root page via header lookup
- [x] get_wiki_headers returns page tree
- [x] Auth error handling
- [x] All 117 tests pass

[DPE-1624]: https://sagebionetworks.jira.com/browse/DPE-1624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Tested the service locally and it successfully gives correct wiki info.
1. Get wikipage 
Screenshot 2026-06-24 at 12 01 19 AM

2. Get wikiheader
Screenshot 2026-06-24 at 12 01 30 AM

3. Get order hint
Screenshot 2026-06-24 at 12 07 52 AM
Screenshot 2026-06-24 at 9 15 04 AM

4. Get wiki page version history
Screenshot 2026-06-24 at 9 17 04 AM


src/synapse_mcp/services/tool_service.py
```diff
@@ -71,6 +71,28 @@ def collect_generator(gen: Iterator, limit: int = 100) -> list:
...
 
+async def collect_async_generator(gen: AsyncIterator, limit: int = 100) -> list:
+    """Collect up to *limit* items from an async generator.
+
+    Async counterpart of ``collect_generator`` for SDK methods that
+    return ``AsyncGenerator`` (e.g. ``WikiHeader.get_async``).
+
+    Raises:
+        ValueError: If ``limit`` is negative.
+    """
```
BryanFauble: This collect_async_generator logic needs to be overhauled. It does not actually provide any way to "paginate" across a particular record. This would only return up to the first 100 results of a generator, and never be able to proceed beyond that to get results 101...1xx

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,166 @@
...
+
+        Returns:
+            Dict with page id, title, markdown content,
+            owner_id, parent_wiki_id, and timestamps.
+        """
+        async with synapse_client(ctx) as client:
+            if wiki_id is None:
+                # SDK requires id or title — find root page
+                # from the wiki header tree.
+                headers = await collect_async_generator(
```
BryanFauble: To swap away from `collect_async_generator`

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,166 @@
...
+                the wiki.
+            offset: Pagination offset (default 0).
+            limit: Maximum headers to return (default 50).
+
+        Returns:
+            List of dicts with id, title, and parent_id
+            for each wiki page in the hierarchy.
+        """
+        async with synapse_client(ctx) as client:
+            headers = await collect_async_generator(
```
BryanFauble: To swap away from collect_async_generator

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,166 @@
...
+            wiki_id: Wiki page ID.
+            offset: Pagination offset (default 0).
+            limit: Max snapshots to return (default 50).
+
+        Returns:
+            List of dicts with version, modified_on,
+            and modified_by for each revision.
+        """
+        async with synapse_client(ctx) as client:
+            snapshots = await collect_async_generator(
```
BryanFauble: To swap away from collect_async_generator

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,201 @@
...
+                            root = header
+                            break
+                except SynapseHTTPError as exc:
+                    if getattr(exc.response, "status_code", None) == 404:
+                        return {
+                            "error": f"No root wiki page found for {owner_id}",
+                            "owner_id": owner_id,
+                        }
+                    raise
+                wiki_id = root.id
```
andrewelamb, line 72: This assumes you find a root (besides getting the SynapseHTTPError). If you don't, you're going to get an error here. What if the owner entity doesn't have any wikis?

tests/test_wiki_service.py
```diff
@@ -0,0 +1,327 @@
...
+    attachment_file_handle_ids: Optional[List[str]] = None
+    
+@dataclass
+class FakeWikiHeader:
+    id: str = "1"
+    title: str = "Home"
+    parent_id: Optional[str] = None
+
+
+class TestGetWikiPage:
```
andrewelamb, line 51: Please add this test, it demonstrates [this](https://github.com/Sage-Bionetworks/synapse-mcp/pull/25/changes#r3469132889).

```
    @patch(f"{TS}.get_synapse_client", new_callable=AsyncMock)
    @patch(f"{SVC}.WikiHeader")
    async def test_given_empty_header_tree_when_fetched_without_id_then_returns_not_found(
        self, mock_wh_cls, mock_get_client
    ):
        mock_get_client.return_value = MagicMock()

        async def _empty(**_):
            if False:
                yield

        mock_wh_cls.get_async = _empty

        result = await WikiService().get_wiki_page(MagicMock(), "syn123")
        # This passes now, but shouldn't
        assert result["error"] == "'NoneType' object has no attribute 'id'"
        # This fails now, but should be the actual result
        assert result["error"] == "No root wiki page found for syn123"
        assert result["owner_id"] == "syn123"
```

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,201 @@
...
+                the wiki.
+            offset: Pagination offset (default 0).
+            limit: Maximum headers to return (default 50).
+
+        Returns:
+            List of dicts with id, title, and parent_id
+            for each wiki page in the hierarchy.
+        """
+        if not owner_id:
+            return {
```
andrewelamb, line 106: This doesn't match the return type

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,201 @@
...
+            wiki_id: The wiki page ID to get the history for.
+            offset: Pagination offset (default 0).
+            limit: Max snapshots to return (default 50).
+
+        Returns:
+            List of dicts with version, modified_on
+            and modified_by for each snapshot in the history.
+        """
+        if not owner_id or not wiki_id:
+            return {
```
andrewelamb, line 158: This doesn't match the return type

src/synapse_mcp/services/tool_service.py
```diff
@@ -8,7 +8,7 @@
 from collections.abc import Mapping
 from contextlib import asynccontextmanager
 from datetime import date, datetime
-from typing import Any, Callable, Dict, Iterator, Tuple
+from typing import Any, AsyncIterator, Callable, Dict, Iterator, Tuple
```
andrewelamb: AsyncIterator doesn't appear to be used

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        if not owner_id:
+            return {
+                "error": ("The owner_id is required to get the wiki order hint"),
+            }
+        async with synapse_client(ctx) as client:
+            hint = await WikiOrderHint(
+                owner_id=owner_id,
+            ).get_async(synapse_client=client)
+            if not hint:
```
andrewelamb: Is this needed? It seems like `WikiOrderHint.get_async` always returns itself. If you intended to check if `WikiOrderHint.id_list` is empty, then this should be 

```
if not hint.id_list
```

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,201 @@
...
+        if not owner_id:
+            return {
+                "error": ("The owner_id is required to get a wiki page"),
+            }
+        async with synapse_client(ctx) as client:
+            if wiki_id is None:
+                # SDK requires id or title — find the root page from the
+                # wiki header tree. Limit the page size to keep root
+                # discovery cheap; we only need the parent_id == None
+                # entry, which lives in the first page.
```
andrewelamb, line 53: Is this strictly correct? Claude suggests this is more accurate:

```
# WikiPage.get requires id or title, so find the root by scanning
# the header tree for the entry whose parent_id is None. get_async
# paginates across the entire tree; limit is only the per-request
# page size, not an overall cap. We break as soon as the root is found.
```

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,201 @@
...
+                offset=offset,
+                limit=limit,
+                synapse_client=client,
+            ):
+                if len(results) >= limit:
+                    break
+                results.append(serialize_model(header))
+            if not results:
+                return {
+                    "error": ("No wiki headers found for the given owner_id"),
```
andrewelamb, line 127: This seems misleading to the user. Shouldn't this be the message only if the offset is 0? 

Say a wiki has 50 headers and the users first call is

`first = get_wiki_headers()`

they would get back all 50, but the user doesn't know that so they make  a second call:

`second = get_wiki_headers(offest=50)`

they would get back:

`{"error": ("No wiki headers found for the given owner_id")}`

tests/test_wiki_service.py
```diff
@@ -0,0 +1,327 @@
...
+
+        assert result["error"] == "The owner_id is required to get the wiki order hint"
+
+    @patch(f"{TS}.get_synapse_client", new_callable=AsyncMock)
+    @patch(f"{SVC}.WikiOrderHint")
+    async def test_given_no_hint_set_when_order_hint_fetched_then_returns_error(
+        self, mock_hint_cls, mock_get_client
+    ):
+        mock_get_client.return_value = MagicMock()
+        mock_hint_cls.return_value.get_async = AsyncMock(return_value=None)
```
andrewelamb: This is a misleading test. See [this comment](https://github.com/Sage-Bionetworks/synapse-mcp/pull/25/changes#r3469274236). This probably should be testing for when the WikiOrderHint.is_list empty.

tests/test_wiki_service.py
```diff
@@ -0,0 +1,327 @@
+"""Tests for WikiService."""
```
andrewelamb, line 1: You might want to add coverage for these use-cases:

1. get_wiki_page — non-404 HTTP error re-raise (wiki_service.py:64-70)
The 404 case is tested (test_given_entity_without_wiki_then_returns_not_found), but the else path — a non-404 SynapseHTTPError (e.g. 403/500) that should be re-raised and handled by the error_boundary — is not tested. Right now nothing exercises the raise on line 70.

2. get_wiki_page — root discovery returns headers but no root page (wiki_service.py:53-71)
The existing test always yields a header with parent_id=None. If the header tree yields entries but none has parent_id is None, root stays None and line 71 (wiki_id = root.id) raises AttributeError. This is both an untested branch and a likely latent bug — worth a test, and possibly a fix to return a clean error instead of crashing.

3. get_wiki_headers — limit truncation (wiki_service.py:119-120)
get_wiki_history has a dedicated truncation test (test_given_limit_when_history_fetched_then_truncates), but the equivalent len(results) >= limit break in get_wiki_headers is never exercised — every headers test yields fewer items than the limit.

4. get_wiki_history — empty result returns []
Unlike get_wiki_headers (which returns an error dict when empty), get_wiki_history just returns an empty list. There's no test asserting that behavior, so the difference in contract between the two methods is untested (and arguably worth confirming it's intentional).

andrewelamb (changes requested):

src/synapse_mcp/services/wiki_service.py
danlu1: If there is no wiki for the owner entity, it should hit the `404 SynapseHTTPError`.

src/synapse_mcp/services/wiki_service.py
danlu1: As the [API doc](https://rest-docs.synapse.org/rest/GET/entity/ownerId/wikiheadertree2.html), "The first WikiHeader will be for the root WikiPage (parentWikiId = null)." so we don't have to exhaust the wiki header tree to get the root wiki.

src/synapse_mcp/services/wiki_service.py
danlu1: I intent to check the id_list. Thanks for catching

src/synapse_mcp/services/wiki_service.py
andrewelamb: So WikiHeader.get_async will never return nothing, and will always contain the root? (or will return a 404 SynapseHTTPError?

src/synapse_mcp/services/wiki_service.py
danlu1: The root page is in the first page so it should be okay.

src/synapse_mcp/services/wiki_service.py
danlu1: It will return a 404 is no wiki page is for the owner entity, see the trackback below. Or it returns the wiki tree with the first WikiHeader as the root WikiPage.

```
>>> list(WikiHeader().get(owner_id = 'syn75938539'))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/core/async_utils.py", line 168, in wrap_async_generator_to_sync_generator
    item = loop.run_until_complete(anext(async_gen))
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/core/async_utils.py", line 160, in async_wrapper
    async for item in async_gen_func(*args, **kwargs):
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/models/wiki.py", line 411, in get_async
    async for item in get_wiki_header_tree(
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/api/wiki_services.py", line 198, in get_wiki_header_tree
    async for item in rest_get_paginated_async(
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/api/api_client.py", line 105, in rest_get_paginated_async
    response = await client.rest_get_async(
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/client.py", line 9560, in rest_get_async
    response = await self._rest_call_async(
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/client.py", line 9530, in _rest_call_async
    self._handle_httpx_synapse_http_error(response)
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/client.py", line 9215, in _handle_httpx_synapse_http_error
    exceptions._raise_for_status_httpx(
  File "/Users/danlu/.pyenv/versions/3.10.11/lib/python3.10/site-packages/synapseclient/core/exceptions.py", line 296, in _raise_for_status_httpx
    raise SynapseHTTPError(message, response=response)
synapseclient.core.exceptions.SynapseHTTPError: 404 Client Error: A root wiki does not exist for ownerId: 75938539 and ownerType: ENTITY
```

Using MCP
Screenshot 2026-06-24 at 2 15 40 PM

tests/test_wiki_service.py
danlu1: As our discussion, this is no need to be added.

tests/test_wiki_service.py
danlu1: Done: get_wiki_page — non-404 HTTP error re-raise (wiki_service.py:64-70)
The 404 case is tested (test_given_entity_without_wiki_then_returns_not_found), but the else path — a non-404 SynapseHTTPError (e.g. 403/500) that should be re-raised and handled by the error_boundary — is not tested. Right now nothing exercises the raise on line 70.

Not applicable since if there is no root page, it will error out with 404 error: get_wiki_page — root discovery returns headers but no root page (wiki_service.py:53-71)
The existing test always yields a header with parent_id=None. If the header tree yields entries but none has parent_id is None, root stays None and line 71 (wiki_id = root.id) raises AttributeError. This is both an untested branch and a likely latent bug — worth a test, and possibly a fix to return a clean error instead of crashing. 

Done. get_wiki_headers — limit truncation (wiki_service.py:119-120)
get_wiki_history has a dedicated truncation test (test_given_limit_when_history_fetched_then_truncates), but the equivalent len(results) >= limit break in get_wiki_headers is never exercised — every headers test yields fewer items than the limit.

Not applicable, since wiki page should always has version history if created: get_wiki_history — empty result returns.
Unlike get_wiki_headers (which returns an error dict when empty), get_wiki_history just returns an empty list. There's no test asserting that behavior, so the difference in contract between the two methods is untested (and arguably worth confirming it's intentional).

andrewelamb (approved):

src/synapse_mcp/services/wiki_service.py
linglp: @danlu1 I agree with @andrewelamb here. The return type is `List[Dict[str, Any]]` but here it only returns a dictionary

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,205 @@
...
+        Arguments:
+            ctx: The FastMCP request context.
+            owner_id: The Synapse ID of the wiki owner.
+
+        Returns:
+            Dict with owner_id, owner_object_type, id_list (page ordering),
+            and etag.
+            Or if the wiki order hint is not set, returns an error.
+        """
+        if not owner_id:
```
linglp, line 193: same as above. this doesn't match the return type

src/synapse_mcp/services/wiki_service.py
```diff
@@ -0,0 +1,205 @@
...
+                    "error": ("No wiki headers found for the given owner_id"),
+                }
+            return results
+
+    @staticmethod
+    @error_boundary(
+        error_context_keys=("owner_id", "wiki_id"),
+        wrap_errors=True,
+    )
+    async def get_wiki_history(
```
linglp, line 136: I think the behaviors of `get_wiki_history` and `get_wiki_page` are inconsistent: 
* get_wiki_page — if no root header is found, returns an explicit error 
* get_wiki_history — if no snapshots exist, silently returns [] (empty list, no error)

tests/test_wiki_service.py
```diff
@@ -0,0 +1,381 @@
...
+        assert result["error"] == "No wiki headers found for the given owner_id"
+
+
+@dataclass
+class FakeWikiHistorySnapshot:
+    version: int = 1
+    modified_on: str = "2025-01-01"
+
+
+class TestGetWikiHistory:
```
linglp, line 251: there's also no test like `test_given_entity_with_no_wiki_then_returns_error` to ensure that the behaviors are expected when there's no wiki history.

linglp: Oh I see! Wiki history is always available! No need to take into consideration of this case then.

src/synapse_mcp/services/wiki_service.py
linglp: No worries, I think in practce, wiki history can't be an empty list

linglp (approved):
LGTM! I also think it worths noting in a comment that every Synapse wiki has exactly one root page `(parent_id=None)` and at least version 0 from the moment it's created (so wiki history is never None)

PR #26, opened 2026-04-09
[DPE-1625] Add Team, User, and Activity services and tools

BryanFauble:
## Summary

Adds 6 tools for team membership, user profiles, and certification checks. PR 4 of 7 in [DPE-1621](https://sagebionetworks.jira.com/browse/DPE-1621).

**Jira:** [DPE-1625](https://sagebionetworks.jira.com/browse/DPE-1625)

## New tools

| Tool | Purpose | SDK Method |
|------|---------|-----------|
| `get_team` | Get team by ID or name | `Team.from_id()` / `Team.from_name()` |
| `get_team_members` | List all members | `Team(id).members()` |
| `get_team_open_invitations` | List pending invitations | `Team(id).open_invitations()` |
| `get_team_membership_status` | Check membership status | `Team(id).get_user_membership_status()` |
| `get_user_profile` | Get profile (by ID, username, or self) | `UserProfile.from_id()` / `.from_username()` / `.get()` |
| `is_user_certified` | Check certification | `UserProfile(id).is_certified()` |

## User profile lookup modes

```mermaid
flowchart TD
    A[get_user_profile] --> B{Arguments?}
    B -->|"user_id=12345"| C["UserProfile.from_id(12345)"]
    B -->|"username='jsmith'"| D["UserProfile.from_username('jsmith')"]
    B -->|"no args"| E["UserProfile().get() → returns self"]
```

## Test plan

- [x] Team lookup by ID and by name, error when neither provided
- [x] User profile by ID, by username, and self
- [x] is_user_certified returns {user_id, is_certified}
- [x] Auth error paths
- [x] All 125 tests pass

[DPE-1625]: https://sagebionetworks.jira.com/browse/DPE-1625?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/synapse_mcp/services/team_service.py
```diff
@@ -0,0 +1,126 @@
...
+                    "error": (
+                        "Either team_id or team_name"
+                        " is required"
+                    )
+                }
+            return serialize_model(team)
+
+    @error_boundary(
+        error_context_keys=("team_id",),
+        wrap_errors=list,
```
philerooski: `wrap_errors=list` is a type error — `error_boundary` declares `wrap_errors: bool = False`, so this should be `wrap_errors=True`. `list` happens to be truthy so it doesn't crash, but it's inconsistent with every other call site in the codebase (`entity_service.py`, `wiki_service.py`, `curation_task_service.py` all use `wrap_errors=True`).

src/synapse_mcp/services/team_service.py
```diff
@@ -0,0 +1,126 @@
...
+
+        Returns:
+            List of team member dicts.
+        """
+        with synapse_client(ctx) as client:
+            team = Team(id=team_id)
+            members = team.members(
+                synapse_client=client,
+            )
+            return [serialize_model(m) for m in members]
```
philerooski: The Jira spec notes that `Team.members()` is a generator, but inspection of the SDK shows that `members_async` actually collects all pages into a full `List` before returning — so this is not a lazy generator and `collect_generator` is not needed here.

The real concern is that for very large teams the SDK silently fetches every member across all API pages before this method returns. There is no way to pass `limit`/`offset` through `Team.members()` — those parameters exist at the `rest_get_paginated_async` level but are not exposed by the high-level model method.

Worth noting this behavior in a comment or in the tool description so callers know the response size is unbounded and entirely determined by team size.

src/synapse_mcp/services/team_service.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @error_boundary(
+        error_context_keys=("team_id",),
+        wrap_errors=list,
```
philerooski: Same issue as the `get_team_members` decorator above — `wrap_errors=list` should be `wrap_errors=True`.

src/synapse_mcp/services/team_service.py
```diff
@@ -0,0 +1,126 @@
...
+        Returns:
+            List of invitation dicts.
+        """
+        with synapse_client(ctx) as client:
+            team = Team(id=team_id)
+            invitations = team.open_invitations(
+                synapse_client=client,
+            )
+            return [
+                serialize_model(i) for i in invitations
```
philerooski: Same note as `get_team_members` above: `open_invitations()` also collects all pages internally — not a lazy generator. No `collect_generator` needed, but the same unbounded response caveat applies.

tests/test_team_service.py
```diff
@@ -0,0 +1,91 @@
...
+            "expired"
+        )
+
+        # WHEN we get a team
+        result = TeamService().get_team(
+            MagicMock(), team_id=100
+        )
+
+        # THEN an auth error is returned
+        assert "Authentication required" in result["error"]
```
philerooski, line 101: This file only has `TestGetTeam`. Three service methods are untested:
- `get_team_members` — no test for success path, empty list, auth error
- `get_team_open_invitations` — same gaps
- `get_team_membership_status` — no tests at all

These follow the same `@error_boundary` + `synapse_client` pattern as `get_team`, so the fixtures are straightforward to add.

tests/test_user_service.py
```diff
@@ -0,0 +1,103 @@
...
+            "expired"
+        )
+
+        # WHEN we get a profile
+        result = UserService().get_user_profile(
+            MagicMock(), user_id=12345
+        )
+
+        # THEN an auth error is returned
+        assert "Authentication required" in result["error"]
```
philerooski, line 113: `is_user_certified` is completely untested — no success path, no auth error path. Should have at least a `TestIsUserCertified` class with a certified/uncertified case and an auth error case, mirroring the pattern in `TestGetUserProfile`.

src/synapse_mcp/tools.py
```diff
@@ -466,6 +468,107 @@ def get_wiki_order_hint(
...
+    return TeamService().get_team(ctx, team_id, team_name)
+
+
+@mcp.tool(
+    title="Get Team Members",
+    description="List all members of a Synapse Team.",
+    annotations=_RO,
+)
+def get_team_members(
+    team_id: int, ctx: Context
```
philerooski: Nit: `ctx` parameter ordering is inconsistent across the 6 new tools. The established convention in this file puts `ctx` **last** (e.g. `get_entity(entity_id, ctx)`, `get_link(entity_id, ctx)`).

`get_team_members`, `get_team_open_invitations`, `get_team_membership_status`, and `is_user_certified` follow the convention, but `get_team` and `get_user_profile` put `ctx` first. All 6 should be consistent — either standardize on `ctx` last to match the rest of the file, or pick one order and apply it uniformly across this PR.

philerooski (commented):
From the DPE-1625 perspective, this PR looks broadly aligned with the Team/User portion of the scope. One thing worth noting is that the ticket describes 8 tools total (including 2 Activity tools), while this PR’s diff mainly adds the 6 Team/User tools. The Activity tools appear to already be present on the base branch, so the story may still be complete overall. I left a few potentially premature inline comments for issues that were discovered by the agent and reviewed by myself.

src/synapse_mcp/services/team_service.py
BryanFauble: We could grab the logic from https://github.com/Sage-Bionetworks/synapsePythonClient/blob/ded460f03d18720345c88a2cc6ac15ba3c942d2d/synapseclient/api/team_services.py#L227-L237 and put it into the MCP server so we can get the generator logic out of the client without needing to push a fix to the SYNPY client to get the behavior.

src/synapse_mcp/services/team_service.py
BryanFauble: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/ded460f03d18720345c88a2cc6ac15ba3c942d2d/synapseclient/api/team_services.py#L296-L306 can be put into this portion of the code with the limit/offset logic

src/synapse_mcp/tools.py
```diff
@@ -446,6 +448,113 @@ async def get_wiki_order_hint(
...
+    ),
+    annotations=_RO,
+)
+async def get_team(
+    ctx: Context,
+    team_id: Optional[int] = None,
+    team_name: Optional[str] = None,
+) -> Dict[str, Any]:
+    """Get a Synapse Team by ID or name."""
+    return await TeamService().get_team(ctx, team_id, team_name)
```
BryanFauble: The class methods can be static so we don't have to create an instance of `TeamService`.

src/synapse_mcp/services/team_service.py
BryanFauble: Fixed — every `@error_boundary` in `team_service.py` now uses `wrap_errors=True`.

src/synapse_mcp/services/team_service.py
BryanFauble: Right, the SDK wrapper exhaustively collects every page. `get_team_members` now bypasses `Team.members_async` entirely and calls `rest_get_paginated_async` against `/teamMembers/{team_id}` directly with caller-supplied `limit` and `offset` so pagination actually reaches the wire. Each returned dict goes through `TeamMember.fill_from_dict` so the response shape stays identical to what the high-level wrapper would have produced. Tool gains an `offset` parameter.

src/synapse_mcp/services/team_service.py
BryanFauble: Same fix — all `@error_boundary` calls in `team_service.py` are `wrap_errors=True` now.

src/synapse_mcp/services/team_service.py
BryanFauble: Same shape as the members fix — `get_team_open_invitations` now calls `rest_get_paginated_async` against `/team/{team_id}/openInvitation` with caller `limit`/`offset`. Tool gains an `offset` parameter.

tests/test_team_service.py
BryanFauble: Backfilled. `tests/test_team_service.py` now has `TestGetTeamMembers` (success, empty, limit cap, negative-limit error, zero-limit short-circuit, auth error), `TestGetTeamOpenInvitations` (same coverage), and `TestGetTeamMembershipStatus` (success + auth error). The pagination tests assert that `limit`/`offset` reach the wire and that responses come back as the typed `TeamMember` shape (snake_case `team_id`/`is_admin`/`member.owner_id`).

tests/test_user_service.py
BryanFauble: Covered — `tests/test_user_service.py::TestIsUserCertified` has three cases: certified user, uncertified user, and auth error.

PR #27, opened 2026-04-09
[DPE-1626] Add Evaluation and Submission services and tools

BryanFauble:
## Summary

Adds 12 tools for DREAM Challenge evaluation queues and submission workflows. PR 5 of 7 in [DPE-1621](https://sagebionetworks.jira.com/browse/DPE-1621).

**Jira:** [DPE-1626](https://sagebionetworks.jira.com/browse/DPE-1626)

### Evaluation (4 tools)
| Tool | Purpose |
|------|---------|
| `get_evaluation` | Get queue by ID or name |
| `list_evaluations` | List queues with filters |
| `get_evaluation_acl` | Get queue ACL |
| `get_evaluation_permissions` | Get user's permissions |

### Submission (8 tools)
| Tool | Purpose |
|------|---------|
| `get_submission` | Get submission by ID |
| `list_evaluation_submissions` | List all submissions to a queue |
| `list_my_submissions` | List current user's submissions |
| `get_submission_count` | Get count (wrapped in dict) |
| `get_submission_status` | Get status + score + annotations |
| `list_submission_statuses` | List statuses for a queue |
| `list_evaluation_submission_bundles` | List submission+status bundles |
| `list_my_submission_bundles` | List user's bundles |

## Key design decisions

```mermaid
flowchart TD
    A[list_evaluations] --> B{Parameters?}
    B -->|"project_id"| C["get_evaluations_by_project()"]
    B -->|"available_only=True"| D["get_available_evaluations()"]
    B -->|"default"| E["get_all_evaluations()"]
    
    F["get_submission_count()"] -->|"SDK returns raw int"| G["Wrap: {evaluation_id, count}"]
    H["Generator methods"] -->|"Use collect_generator"| I["Cap at limit items"]
```

## Test plan

- [x] All 125 tests pass

[DPE-1626]: https://sagebionetworks.jira.com/browse/DPE-1626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/synapse_mcp/services/evaluation_service.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        async with synapse_client(ctx) as client:
+            ev = Evaluation(id=evaluation_id)
+            perms = await ev.get_permissions_async(
+                synapse_client=client,
+            )
+            return {
+                "evaluation_id": evaluation_id,
+                "permissions": serialize_model(perms),
+            }
```
BryanFauble, line 157: To determine: Is ACL and Permisisons the "same thing" for Evaluations, ie: Can we skip needing both.

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,167 @@
...
+    ) -> List[Dict[str, Any]]:
+        """List all submissions to an Evaluation."""
+        async with synapse_client(ctx) as client:
+            gen = Submission.get_evaluation_submissions_async(
+                evaluation_id=evaluation_id,
+                status=status,
+                synapse_client=client,
+            )
+            items = await collect_async_generator(gen, limit)
+            return [serialize_model(s) for s in items]
```
BryanFauble: Using these `collect_async_generator` is actually incorrect, we will need to expose a limit AND offset for these.

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,167 @@
...
+        evaluation_id: str,
+        limit: int = 50,
+    ) -> List[Dict[str, Any]]:
+        """List the current user's submissions."""
+        async with synapse_client(ctx) as client:
+            gen = Submission.get_user_submissions_async(
+                evaluation_id=evaluation_id,
+                synapse_client=client,
+            )
+            items = await collect_async_generator(gen, limit)
```
BryanFauble: To swap away from `collect_async_generator`

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,167 @@
...
+        """List submission+status bundles for an Evaluation."""
+        async with synapse_client(ctx) as client:
+            gen = (
+                SubmissionBundle.get_evaluation_submission_bundles_async(
+                    evaluation_id=evaluation_id,
+                    status=status,
+                    synapse_client=client,
+                )
+            )
+            items = await collect_async_generator(gen, limit)
```
BryanFauble: To swap away from `collect_async_generator`

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,167 @@
...
+    ) -> List[Dict[str, Any]]:
+        """List current user's submission bundles."""
+        async with synapse_client(ctx) as client:
+            gen = (
+                SubmissionBundle.get_user_submission_bundles_async(
+                    evaluation_id=evaluation_id,
+                    synapse_client=client,
+                )
+            )
+            items = await collect_async_generator(gen, limit)
```
BryanFauble: To swap away from `collect_async_generator`

src/synapse_mcp/tools.py
```diff
@@ -555,6 +557,251 @@ async def is_user_certified(
...
+    ),
+    annotations=_RO,
+)
+async def get_evaluation_permissions(
+    evaluation_id: str, ctx: Context
+) -> Dict[str, Any]:
+    """Get permissions on an Evaluation queue."""
+    return await EvaluationService().get_evaluation_permissions(
+        ctx, evaluation_id
+    )
```
BryanFauble, line 656: One of these might go away if they do not have different meanings.

src/synapse_mcp/tools.py
```diff
@@ -555,6 +557,251 @@ async def is_user_certified(
...
+        "queue, optionally filtered by status. Up to "
+        "``limit`` submissions are returned."
+    ),
+    annotations=_RO,
+)
+async def list_evaluation_submissions(
+    evaluation_id: str,
+    ctx: Context,
+    status: Optional[str] = None,
+    limit: int = 50,
```
BryanFauble, line 693: This would need a limit and offset

src/synapse_mcp/tools.py
```diff
@@ -555,6 +557,251 @@ async def is_user_certified(
...
+    description=(
+        "List the current user's submissions to a "
+        "Synapse Evaluation queue."
+    ),
+    annotations=_RO,
+)
+async def list_my_submissions(
+    evaluation_id: str,
+    ctx: Context,
+    limit: int = 50,
```
BryanFauble, line 714: This would need to have a limit and offset.

src/synapse_mcp/tools.py
```diff
@@ -555,6 +557,251 @@ async def is_user_certified(
...
+        "List submission+status bundles for a Synapse "
+        "Evaluation queue."
+    ),
+    annotations=_RO,
+)
+async def list_evaluation_submission_bundles(
+    evaluation_id: str,
+    ctx: Context,
+    status: Optional[str] = None,
+    limit: int = 50,
```
BryanFauble, line 792: Need limit and offset.

src/synapse_mcp/tools.py
```diff
@@ -555,6 +557,251 @@ async def is_user_certified(
...
+    description=(
+        "List the current user's submission bundles "
+        "for a Synapse Evaluation queue."
+    ),
+    annotations=_RO,
+)
+async def list_my_submission_bundles(
+    evaluation_id: str,
+    ctx: Context,
+    limit: int = 50,
```
BryanFauble, line 813: Needs a limit and offset

src/synapse_mcp/services/evaluation_service.py
```diff
@@ -0,0 +1,153 @@
...
+
+        Arguments:
+            ctx: The FastMCP request context.
+            evaluation_id: Evaluation queue ID.
+
+        Returns:
+            Dict with ACL information.
+        """
+        async with synapse_client(ctx) as client:
+            ev = Evaluation(id=evaluation_id)
```
jaymedina, line 128: You're not actually retrieving the Evaluation object here(?) (no `.get_evaluation`)

Also, the python client lets you retrieve the ACL from an Evaluation you got via `name` as well:

Image

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,269 @@
...
+        wrap_errors=True,
+    )
+    async def list_my_submissions(
+        ctx: Context,
+        evaluation_id: str,
+        offset: int = 0,
+        limit: int = 50,
+    ) -> List[Dict[str, Any]]:
+        """List the caller's own submissions to an Evaluation."""
+        if limit < 0 or offset < 0:
```
jaymedina: nit: I see these lines repeated in all the SubmissionService methods. can we abstract this away for a bit of cleanup?

tests/test_submission_service.py
```diff
@@ -0,0 +1,263 @@
...
+            captured["limit"] = limit
+            captured["offset"] = offset
+            for sid in ("1", "2"):
+                yield {"id": sid, "evaluationId": "9600001"}
+
+        mock_paginate.side_effect = _fake
+
+        # WHEN we list submissions with a status filter and offset
+        result = await SubmissionService.list_evaluation_submissions(
+            MagicMock(), "9600001", status="SCORED", offset=10, limit=2
```
jaymedina, line 112: do you want to test the `ValueError` too, and try a case where limit/offset < 0?

jaymedina (approved):
LGTM! Just some small comments here and there.

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,269 @@
...
+        )
+        async with synapse_client(ctx) as client:
+            results: List[Dict[str, Any]] = []
+            async for raw in rest_get_paginated_async(
+                uri=uri,
+                limit=limit,
+                offset=offset,
+                synapse_client=client,
+            ):
+                if len(results) >= limit:
```
andrewelamb: Wouldn't this check make sense after the append on line 101? It would be one less trip through the loop.

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,269 @@
...
```
andrewelamb: Same as above

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,269 @@
...
```
andrewelamb: same as above

src/synapse_mcp/services/submission_service.py
```diff
@@ -0,0 +1,269 @@
...
```
andrewelamb: same as above

src/synapse_mcp/services/submission_service.py
andrewelamb: This pattern seems to be repeated several times below. Perhaps it should be separated into a function?

src/synapse_mcp/services/evaluation_service.py
```diff
@@ -0,0 +1,153 @@
...
+        async with synapse_client(ctx) as client:
+            if evaluation_id is not None:
+                ev = await Evaluation(id=evaluation_id).get_async(
+                    synapse_client=client,
+                )
+            elif evaluation_name is not None:
+                ev = await Evaluation(name=evaluation_name).get_async(
+                    synapse_client=client,
+                )
+            else:
```
andrewelamb: This if/elif/else block should happen outside the login to synapse block

tests/test_evaluation_service.py
```diff
@@ -0,0 +1,180 @@
...
+        result = await EvaluationService().get_evaluation(
+            MagicMock(), evaluation_name="Named Eval"
+        )
+
+        assert result["name"] == "Named Eval"
+
+    async def test_given_no_args_then_returns_error(self):
+        result = await EvaluationService().get_evaluation(MagicMock())
+
+        assert "required" in result["error"]
```
andrewelamb, line 81: Are we sure this is testing the right thing? I assume this is looking for the  "Either evaluation_id or evaluation_name is required" error message, but I'm worrried that the sctual error message here  is ""Authentication required: ...", and this is s false positive.

src/synapse_mcp/services/evaluation_service.py
```diff
@@ -0,0 +1,153 @@
...
+        ctx: Context,
+        project_id: Optional[str] = None,
+        access_type: Optional[str] = None,
+        active_only: Optional[bool] = None,
+        available_only: bool = False,
+        evaluation_ids: Optional[List[str]] = None,
+        offset: int = 0,
+        limit: int = 20,
+    ) -> List[Dict[str, Any]]:
+        """List evaluations with optional filters.
```
andrewelamb, line 66: Some filters aren't used in every branch in the function, for example, `access_type` isn't used in the function call on line 91. Should this be in the docstring?

src/synapse_mcp/services/evaluation_service.py
```diff
@@ -0,0 +1,153 @@
...
+    @error_boundary(
+        error_context_keys=("evaluation_id", "evaluation_name")
+    )
+    async def get_evaluation(
+        ctx: Context,
+        evaluation_id: Optional[str] = None,
+        evaluation_name: Optional[str] = None,
+    ) -> Dict[str, Any]:
+        """Get an Evaluation by ID or name.
+
```
andrewelamb, line 24: nit: The docstring should mention what happens when both id and name are supplied

src/synapse_mcp/services/evaluation_service.py
```diff
@@ -0,0 +1,153 @@
...
+                return {
+                    "error": (
+                        "Either evaluation_id or "
+                        "evaluation_name is required"
+                    )
+                }
+            return serialize_model(ev)
+
+    @staticmethod
+    @error_boundary(wrap_errors=True)
```
andrewelamb, line 55: Should this have error context keys?

tests/test_evaluation_service.py
```diff
@@ -0,0 +1,180 @@
+"""Tests for EvaluationService."""
```
andrewelamb, line 1: You might want to add tests for these scenarios:

1. Filter & pagination passthrough in list_evaluations (biggest gap)
Every test just asserts assert_called_once() — none verify the arguments actually get forwarded. So access_type, active_only, evaluation_ids, offset, and limit are completely untested. Given the recent "REST pagination" work (commit c1a3aeb), this is the most important hole. You'd want assert_called_once_with(...) checking those kwargs reach get_all_evaluations_async / get_evaluations_by_project_async / get_available_evaluations_async.

2. Branch precedence
- get_evaluation: when both evaluation_id and evaluation_name are passed, ID wins. Untested.
- list_evaluations: project_id takes priority over available_only. Untested — a test passing both should confirm the project-scoped path runs.

3. No error handling at all for get_evaluation_acl and get_evaluation_permissions
Both only have a happy-path test. Neither has an auth-failure test, unlike get_evaluation/list_evaluations. These use error_context_keys=("evaluation_id",), so a failure should return an error dict that includes evaluation_id — worth asserting.

4. error_context_keys enrichment is never asserted
The auth-error tests check the "error" message but never that evaluation_id/evaluation_name are echoed back into the error dict (lines 197–209 of tool_service.py). That's the whole point of error_context_keys and it's untested for this service.

5. Generic (non-auth) exception path + status_code surfacing
Only ConnectionAuthError is exercised. The except Exception branch — including the status_code extraction from exc.response (lines 212–229) — has no coverage here. A test raising e.g. a SynapseHTTPError-shaped exception would confirm status_code lands in the result.

6. Minor
- Empty-list result from list_evaluations (returns []).
- get_evaluation_acl asserts "acl" in result but doesn't assert the serialized ACL contents — weak assertion.

tests/test_submission_service.py
```diff
@@ -0,0 +1,263 @@
+"""Tests for SubmissionService."""
```
andrewelamb, line 1: Some issues:

- No tests for list_my_submissions
- All four paginated methods share guards that no test touches

andrewelamb (changes requested):

src/synapse_mcp/services/evaluation_service.py
andrewelamb: Agreed!

src/synapse_mcp/services/submission_service.py
andrewelamb: Agreed!

tests/test_submission_service.py
andrewelamb: Agreed!

src/synapse_mcp/services/evaluation_service.py
jaymedina: I'm actually going to backpedal on what I said here. The `get_acl()` method for the Evaluation model requires the `id` attribute to work, not `name`. See screenshot below:

image

----

And here is the attempt with `name`:

image

[REDACTED]

image

----

We _could_ do something like this, so that users can use the MCP to retrieve their evaluations' ACLs from not just `id` but also `name`:

image


But that kind of quality of life improvement should happen at the **root** (i.e. in `get_acl` itself on SYNPY), not here.

@andrewelamb LMK what you think before I resolve this.

andrewelamb: > The get_acl() method for the Evaluation model requires the id attribute to work, not name

If that's the case then go ahead and resolve this comment

src/synapse_mcp/services/submission_service.py
jaymedina: Implemented! https://github.com/Sage-Bionetworks/synapse-mcp/pull/27/commits/2707b9584fb7b9e3ea6047d57a32edffd8c6d0a6

src/synapse_mcp/services/submission_service.py
jaymedina: 👍 Implemented! https://github.com/Sage-Bionetworks/synapse-mcp/pull/27/commits/2707b9584fb7b9e3ea6047d57a32edffd8c6d0a6

src/synapse_mcp/services/evaluation_service.py
jaymedina: Implemented! https://github.com/Sage-Bionetworks/synapse-mcp/pull/27/commits/63464b361449dbe00cf4265ce1705e87a4d65729

tests/test_evaluation_service.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Good catch — this was a valid concern. The guard has been moved outside the `async with synapse_client` block (https://github.com/Sage-Bionetworks/synapse-mcp/pull/27/commits/63464b361449dbe00cf4265ce1705e87a4d65729), so auth is never attempted when both args are missing. The test is now unambiguously correct: no auth mock is needed and the error will always be the validation message.

src/synapse_mcp/services/evaluation_service.py
jaymedina: https://github.com/Sage-Bionetworks/synapse-mcp/pull/27/commits/c87e1ecbd482145dd61a858c2667f22088799fd3

src/synapse_mcp/services/evaluation_service.py
jaymedina: I'm seeing a pattern in the script: `error_context_keys` are only applied for non-optional args. Not sure how useful it would be in the logs for optional args to be passed as error context keys.

I vote to leave this out.

andrewelamb (approved):

PR #31, opened 2026-04-10
[DPE-1633] Implement load_one method in ClientRegistry and its subclasses for single client retrieval

BryanFauble:
# **Problem:**

  - There was a recent update to the FastMCP server that caused a regression on container restarts. After a container restart, previously registered OAuth clients become invisible to the get_client() lookup. Root cause: FastMCP 2.14.6 stores clients in _client_store (an encrypted DiskStore), but SessionAwareOAuthProxy._restore_registered_clients() writes to self._clients (a dict that FastMCP no longer reads). The DiskStore is ephemeral in Fargate containers, so after restart the clients are in Redis but never loaded into _client_store.


# **Solution:**

  - Change to load a single client from redis on demand instead of reloading all registered clients on startup of the container. The implication of this change is more traffic and communication with redis, however, the benefit is that the memory does not grow with more client registrations being added to the system, not every client is loaded back. Instead, clients are loaded back into memory as they communicate with the MCP server again.                                                                                                                                                       
  - Removed session-storage/token-mirroring code. The previous implementation stored user tokens in a separate session store that was coupled to the client registry restore path. With the move to on-demand client loading, that code and its associated tests became dead weight and have been removed.    

```mermaid
sequenceDiagram                                                                                                                                                                                    
      participant Client as MCP Client                                                                                                                                                                 
      participant WAF as AWS WAF                                                                                                                                                                       
      participant Proxy as SessionAwareOAuthProxy                                                                                                                                                      
      participant Store as _client_store<br/>(DiskStore, ephemeral)                                                                                                                                    
      participant Redis as Redis<br/>(persistent)           
      participant IdP as Synapse IdP                                                                                                                                                                   
                                                            
      Note over Client,IdP: REGISTRATION PHASE                                                                                                                                                         
                                                            
      Client->>WAF: POST /register                                                                                                                                                                     
      WAF->>Proxy: POST /register                           
      activate Proxy
      Proxy->>Proxy: register_client()                                                                                                                                                                 
      Proxy->>Store: super().register_client()<br/>put(client_id, ProxyDCRClient)
      Store-->>Proxy: stored ✓                                                                                                                                                                         
      Proxy->>Redis: _client_registry.save()<br/>hset(client_id, registration)
      Redis-->>Proxy: stored ✓                                                                                                                                                                         
      Proxy-->>Client: 201 Created                          
      deactivate Proxy                                                                                                                                                                                 
                                                            
      Note over Client,IdP: AUTHORIZE (same container — DiskStore hit)                                                                                                                                 
                                                            
      Client->>WAF: GET /authorize?client_id=X                                                                                                                                                         
      WAF->>Proxy: GET /authorize                           
      activate Proxy                                                                                                                                                                                   
      Proxy->>Store: super().get_client(X)                  
      Store-->>Proxy: ProxyDCRClient ✓                                                                                                                                                                 
      Note right of Proxy: Redis NOT consulted
      Proxy->>Proxy: authorize() → create transaction                                                                                                                                                  
      Proxy-->>Client: 302 → /consent                       
      deactivate Proxy                                                                                                                                                                                 
                                                            
      rect rgb(255, 240, 240)                                                                                                                                                                          
          Note over Store: ══ CONTAINER RESTARTS ══<br/>DiskStore WIPED<br/>Redis persists
      end                                                                                                                                                                                              
  
      Note over Client,IdP: AUTHORIZE (new container — Redis fallback + cache)                                                                                                                         
                                                            
      Client->>WAF: GET /authorize?client_id=X                                                                                                                                                         
      WAF->>Proxy: GET /authorize                           
      activate Proxy
      Proxy->>Store: super().get_client(X)
      Store-->>Proxy: None (wiped)
      Proxy->>Redis: load_one(X) → hget                                                                                                                                                                
      Redis-->>Proxy: ClientRegistration ✓                                                                                                                                                             
      Proxy->>Proxy: _registration_to_proxy_client()                                                                                                                                                   
      Proxy->>Store: _client_store.put(X, ProxyDCRClient)                                                                                                                                              
      Store-->>Proxy: cached ✓                              
      Proxy-->>Client: 302 → /consent                                                                                                                                                                  
      deactivate Proxy                                      
                                                                                                                                                                                                       
      Client->>WAF: GET /consent                            
      WAF->>Proxy: GET /consent
      activate Proxy
      Proxy->>Store: super().get_client(X)                                                                                                                                                             
      Store-->>Proxy: ProxyDCRClient ✓ (cached)
      Note right of Proxy: Redis NOT consulted                                                                                                                                                         
      Proxy-->>Client: 200 consent HTML                                                                                                                                                                
      deactivate Proxy
                                                                                                                                                                                                       
      Client->>IdP: GET /authorize (upstream)                                                                                                                                                          
      IdP-->>Client: 302 → callback
      Client->>Proxy: GET /oauth/callback?code=...                                                                                                                                                     
      Proxy->>IdP: POST /token (upstream exchange)                                                                                                                                                     
      IdP-->>Proxy: {access_token, refresh_token}
      Proxy->>Proxy: _handle_idp_callback()<br/>stores code in _code_store                                                                                                                             
      Proxy-->>Client: 302 → redirect_uri?code=LOCAL_CODE                                                                                                                                              
                                                                                                                                                                                                       
      Client->>WAF: POST /token (client_id=X, code=...)                                                                                                                                                
      WAF->>Proxy: POST /token                              
      activate Proxy                                                                                                                                                                                   
      Proxy->>Proxy: ClientAuthenticator.authenticate_request()
      Proxy->>Store: super().get_client(X)                                                                                                                                                             
      Store-->>Proxy: ProxyDCRClient ✓ (cached)
      Note right of Proxy: Redis NOT consulted                                                                                                                                                         
      Proxy->>Proxy: exchange_authorization_code()<br/>FastMCP issues its own JWT
      Proxy-->>Client: {access_token (JWT), refresh_token}                                                                                                                                             
      deactivate Proxy        
```

# **Testing:**

- Unit testing the changes, and will deploy to both dev and staging to verify changes.

### DIFF
--- a/src/synapse_mcp/oauth/proxy.py
+++ b/src/synapse_mcp/oauth/proxy.py
@@ -25,6 +25,7 @@ class SessionAwareOAuthProxy(OAuthProxy):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self._client_registry = create_client_registry(os.environ)
+        self._static_registrations = load_static_registrations()
         logger.debug(
             "SessionAwareOAuthProxy initialized with client registry %s",
             type(self._client_registry).__name__,
@@ -61,11 +62,13 @@ class SessionAwareOAuthProxy(OAuthProxy):
 
         # Persistent registry lookup (Redis hget or file read)
         registration = self._client_registry.load_one(client_id)
+        source = "persistent registry"
         if registration is None:
-            # Also check static registrations
-            for static in load_static_registrations():
+            # Also check static registrations (cached at init)
+            for static in self._static_registrations:
                 if static.client_id == client_id:
                     registration = static
+                    source = "static registrations"
                     break
         if registration is None:
             return None
@@ -80,7 +83,7 @@ class SessionAwareOAuthProxy(OAuthProxy):
         except Exception as exc:  # pragma: no cover - defensive
             logger.debug("Failed to cache client %s in local store: %s", client_id, exc)
 
-        logger.info("Resolved client %s from persistent registry", client_id)
+        logger.info("Resolved client %s from %s", client_id, source)
         return proxy_client
 
     async def register_client(self, client_info):
--- a/src/synapse_mcp/oauth/proxy.py
+++ b/src/synapse_mcp/oauth/proxy.py
@@ -75,14 +75,6 @@ class SessionAwareOAuthProxy(OAuthProxy):
 
         proxy_client = self._registration_to_proxy_client(registration)
 
-        # Cache into _client_store so subsequent lookups in this
-        # container's lifetime (e.g. /consent, /token in the same
-        # auth flow) hit the local store instead of Redis.
-        try:
-            await self._client_store.put(key=client_id, value=proxy_client)
-        except Exception as exc:  # pragma: no cover - defensive
-            logger.debug("Failed to cache client %s in local store: %s", client_id, exc)
-
         logger.info("Resolved client %s from %s", client_id, source)
         return proxy_client
 
--- a/src/synapse_mcp/oauth/proxy.py
+++ b/src/synapse_mcp/oauth/proxy.py
@@ -10,7 +10,6 @@ from fastmcp.server.auth.oauth_proxy import ProxyDCRClient
 from mcp.server.auth.provider import OAuthClientInformationFull
 from pydantic import AnyUrl, TypeAdapter
 
-from ..session_storage import create_session_storage
 from .client_registry import (
     ClientRegistration,
     create_client_registry,
@@ -21,15 +20,13 @@ logger = logging.getLogger("synapse_mcp.oauth")
 
 
 class SessionAwareOAuthProxy(OAuthProxy):
-    """OAuth proxy that mirrors tokens into session storage."""
+    """OAuth proxy with persistent client registry backed by Redis/file."""
 
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
-        self._session_storage = create_session_storage()
         self._client_registry = create_client_registry(os.environ)
         logger.debug(
-            "SessionAwareOAuthProxy initialized with session storage %s and client registry %s",
-            type(self._session_storage).__name__,
+            "SessionAwareOAuthProxy initialized with client registry %s",
             type(self._client_registry).__name__,
         )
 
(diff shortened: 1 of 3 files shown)


anngvu (approved):
OK, so this is a better restore flow, especially for restart-heavy situations. I haven't kept up with all of FastMCP's changes, exactly, but I have a better understanding of how our current implementation relies too much on keeping up with FastMCP's internals so we know where to write or check clients (the ultimate source of this may be me). This seems like the right fix, but ultimately need better maintainability/integration testing for future FastMCP changes (as a followup).

src/synapse_mcp/oauth/proxy.py
```diff
@@ -20,60 +20,71 @@
...
-            try:
-                adapter = TypeAdapter(List[AnyUrl])
-                redirect_source = record.redirect_uris if record.redirect_uris else [
-                    "http://127.0.0.1"]
-                redirect_uris = adapter.validate_python(redirect_source)
-                proxy_client = ProxyDCRClient(
-                    client_id=record.client_id,
-                    client_secret=record.client_secret,
-                    redirect_uris=redirect_uris,
-                    grant_types=record.grant_types or default_grants,
```
BryanFauble, line 66: `SessionAwareOAuthProxy` extends `OAuthProxy` from FastMCP. In FastMCP <2.14.6, `OAuthProxy.get_client()` looked up registered clients from `self._clients` (a plain `dict`). Our `_restore_registered_clients()` method loaded persisted clients from Redis and wrote them back into `self._clients` — this worked fine.

In FastMCP 2.14.6, `OAuthProxy` switched its internal storage to `self._client_store` (an encrypted `DiskStore`). `get_client()` now reads from `_client_store`, not `_clients`. Since `_client_store` is ephemeral (wiped on Fargate container restart), and our restore logic was still writing to the now-ignored `_clients` dict, every previously registered client became invisible after a restart.

This is a consequence of depending on an internal attribute (`_clients`) of a third-party class — when the upstream implementation changed, our subclass silently broke.

src/synapse_mcp/oauth/proxy.py
```diff
@@ -20,60 +20,71 @@
...
+        # container's lifetime (e.g. /consent, /token in the same
+        # auth flow) hit the local store instead of Redis.
         try:
-            registrations.extend(load_static_registrations())
+            await self._client_store.put(key=client_id, value=proxy_client)
         except Exception as exc:  # pragma: no cover - defensive
-            logger.warning("Failed to load static OAuth clients: %s", exc)
-
-        default_grants = ["authorization_code", "refresh_token"]
+            logger.debug("Failed to cache client %s in local store: %s", client_id, exc)
```
BryanFauble: The previous `_restore_registered_clients()` ran during `__init__` and called `load_all()`, which pulled **every** registered client from Redis into memory at once. In practice, only a small fraction of registered clients are active at any given time. This meant:

- Startup latency scaled with the total number of ever-registered clients
- Memory usage grew unbounded as more clients registered over time
- Every container restart paid the full cost even if most clients never reconnected

The new approach loads clients on demand via `load_one()` and caches them in `_client_store` (the correct store). Only clients that actually make requests get loaded, and the cache is naturally bounded to the active working set.

BryanFauble (commented):
Added inline context explaining the `_clients` → `_client_store` breaking change and the motivation for replacing bulk restore with on-demand loading.

src/synapse_mcp/oauth/proxy.py
```diff
@@ -20,60 +20,71 @@
...
```
BryanFauble: This code should be removed in the event that multiple containers need to keep these in sync. The source of truth should be the redis instance.

PR #32, opened 2026-04-10
[DPE-1633] Add Redis storage creation for OAuth state with encryption support

BryanFauble:
**Problem:**
                                                                                                                                                                                                       
  When the ECS container restarts (deployment, scaling, crash recovery), all authenticated users are forced to re-authenticate through the full Synapse OAuth browser flow. This happens because FastMCP 2.14.6 stores all OAuth state — upstream tokens, JTI mappings, refresh tokens, authorization codes — in an encrypted DiskStore. In Fargate, the container's disk is ephemeral and wiped on every restart.
                                                                                                                                                                                                       
The previous token-mirroring approach (removed in #31) relied on accessing self._access_tokens, a plain Python dict that FastMCP 2.14.6 replaced with opaque PydanticAdapter collections backed by the DiskStore. After the upgrade, getattr(self, "_access_tokens", {}) silently returned an empty dict, making the mirroring a no-op — tokens were never actually being persisted to Redis. PR #31correctly identified this as dead code and removed it, but did not replace it with the correct integration point.                                                                                    


  **Migration Impact:**                                                                                                                                                                                    
                                                                                                                                                                                                       
  Deploying this change requires a one-time re-authentication for all active users. Existing sessions cannot be preserved across the deploy because:
                                                                                                                                                                                                       
  1. Old session data (synapse_mcp:session:user:{subject} keys) stores simple user_subject → access_token mappings                                                                                     
  2. FastMCP's internal stores expect encrypted UpstreamTokenSet, JTIMapping, and RefreshTokenMetadata objects in mcp-*:: prefixed keys                                                                
  3. The data structures are incompatible and there is no migration path between them                                                                                                                  
                                                                                                                                                                                                       
  What happens on deploy:                                                                                                                                                                              
  - ECS rolls out new containers — the old ephemeral DiskStore is wiped                                                                                                                                
  - The new Redis-backed store starts empty (different key namespace)                                                                                                                                  
  - Old synapse_mcp:session:* keys in Redis are orphaned (nothing reads them)
  - All users re-authenticate once through the Synapse OAuth flow                                                                                                                                      
  - After that, sessions survive future container restarts         

  **Solution:**

  FastMCP's OAuthProxy accepts a client_storage: AsyncKeyValue constructor parameter. When provided, all internal stores (_upstream_token_store, _jti_mapping_store, _refresh_token_store, _code_store, _client_store) use this backend instead of the default ephemeral DiskStore.
                                                                                                                                                                                                       
  The fix passes a Redis-backed RedisStore (from the key_value library, already a transitive dependency via FastMCP) wrapped with FernetEncryptionWrapper as client_storage. This makes all OAuth state persist in Redis automatically — no custom mirroring logic needed. When REDIS_URL is not set (local dev), behavior is unchanged.
                                                                                                                                                                                                       
  Files changed:                                            
  - src/synapse_mcp/oauth/factory.py — Added _create_redis_storage() and pass result as client_storage to SessionAwareOAuthProxy
  - tests/test_oauth_factory.py — New tests for Redis storage creation and fallback behavior                                                                                                           
  
  **Testing:**                                                                                                                                                                                             
                                                            
  - All 63 existing tests pass with no changes                                                                                                                                                         
  - New unit tests verify:                                  
    - _create_redis_storage() returns None without REDIS_URL (DiskStore fallback)                                                                                                                      
    - _create_redis_storage() returns FernetEncryptionWrapper when REDIS_URL is set                                                                                                                    
  - Deploy to dev and verify:                                                                                                                                                                          
    - Authenticate via OAuth                                                                                                                                                                           
    - Restart ECS container                                                                                                                                                                            
    - Confirm MCP client resumes without re-authentication  
    - Verify token keys appear in Redis with appropriate TTLs

### DIFF
--- a/src/synapse_mcp/oauth/factory.py
+++ b/src/synapse_mcp/oauth/factory.py
@@ -2,11 +2,8 @@
 
 import logging
 import os
+from collections.abc import Mapping
 from typing import Optional
-
-from cryptography.fernet import Fernet
-from fastmcp.server.auth.jwt_issuer import derive_jwt_key
-from key_value.aio.wrappers.encryption import FernetEncryptionWrapper
 
 from .config import load_oauth_settings, should_skip_oauth
 from .jwt import SynapseJWTVerifier
@@ -95,7 +92,7 @@ def create_oauth_proxy(env: Optional[dict[str, str]] = None):
     return auth
 
 
-def _create_redis_storage(env: dict[str, str], client_secret: str):
+def _create_redis_storage(env: Mapping[str, str], client_secret: str):
     """Create a Redis-backed encrypted storage for OAuth state.
 
     When REDIS_URL is available, returns a FernetEncryptionWrapper around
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 This makes sense - does Synapse oauth set a time limit on the secrets returned? Meaning, will users be stuck in an internal loop if their token is expired or will they be able to re-authenticate?

BryanFauble:
> 🔥 This makes sense - does Synapse oauth set a time limit on the secrets returned? Meaning, will users be stuck in an internal loop if their token is expired or will they be able to re-authenticate?

No. Both the refresh token should also work. If that is expired it should put users on the flow to re auth through the normal flow @thomasyu888

PR #36, opened 2026-04-12
[DPE-1635] Remove redundant ClientRegistry and dead session_storage

thomasyu888:
# **Problem:**

- After upgrading to FastMCP 3.2.3 (via PRs #31 and #32), two separate Redis-backed stores were persisting OAuth client registrations simultaneously:
  - FastMCP's built-in `client_storage` (Fernet-encrypted, configured in `factory.py`) — the authoritative store
  - The custom `ClientRegistry` (`FileClientRegistry` / `RedisClientRegistry` in `client_registry.py`) — an unencrypted, redundant second copy
- The `ClientRegistry` was the correct solution when originally written (Sept 2025) against FastMCP 2.12.x, which had no built-in persistence hook. PR #32 wired up `client_storage` but did not remove the now-redundant custom registry, leaving both active.
- Separately, the entire `session_storage` package (`InMemorySessionStorage`, `RedisSessionStorage`, etc.) had no callers anywhere in production code — it was dead code accumulating maintenance surface.
- A static client fallback (`SYNAPSE_MCP_STATIC_CLIENTS`) was also carried forward, but all MCP clients (Claude, OpenAI) use Dynamic Client Registration natively and the env var was never set in any deployment.

# **Solution:**

- Deleted `client_registry.py` (~285 lines: `ClientRegistry`, `FileClientRegistry`, `RedisClientRegistry`, `create_client_registry`).
- Deleted the entire `session_storage/` package (4 files, ~540 lines) and its smoke test script.
- Removed the static client machinery (`_StaticClient`, `_load_static_clients`, `_static_to_proxy_client`) — unused by any deployed MCP client.
- Simplified `proxy.py` to just the grant type normalization in `register_client` and the `state=None` sanitization in `_handle_idp_callback`. `get_client` is now inherited directly from `OAuthProxy`.
- FastMCP's `client_storage` (Redis-backed, Fernet-encrypted, configured in `factory.py`) remains untouched as the sole persistence layer — no change to memory-leak protection or cross-restart durability.

# **Testing:**

- Removed `test_client_registry.py`, `test_session_storage.py`, `test_session_storage_memory.py`, `test_session_storage_redis.py` (all tested deleted code).
- Updated `test_oauth_proxy.py` to remove tests that relied on `FakeRegistry`, the file-based persistence path, and static client loading.
- Full test suite passes: 48 passed, 0 failed.


thomasyu888 (commented):
Sept 2025: ClientRegistry introduced — the only persistence layer at the time
April 10, 2026 (PR #32): client_storage wired into factory.py — from this point, both stores were written on every register_client call
This fix: removes the ClientRegistry read path entirely
Any MCP client that registered before April 10 and hasn't re-registered since is stored only in the old Redis hash (synapse_mcp:client_registry) — not in FastMCP's encrypted store (synapse-mcp-oauth). After this fix goes to production, get_client won't find them and they'll get auth errors, forcing re-authentication.

The other risks are minor by comparison:

* The orphaned synapse_mcp:client_registry Redis keys will just sit there wasting a trivial amount of memory — not harmful, but worth a one-time cleanup
* Tighter coupling to FastMCP's storage schema (the original ticket flagged this as a known con)

This requires strict review

PR #29, opened 2026-04-09
[DPE-1628] Add Utility and Docker services (4.12.0 new features)

BryanFauble:
## Summary

Adds 4 tools using new synapseclient 4.12.0 operations and the DockerRepository model. This is the final PR (7 of 7) in [DPE-1621](https://sagebionetworks.jira.com/browse/DPE-1621).

**Jira:** [DPE-1628](https://sagebionetworks.jira.com/browse/DPE-1628)

## New tools

| Tool | Purpose | SDK Function |
|------|---------|-------------|
| `find_entity_id` | Find entity ID by name + parent | `operations.find_entity_id()` |
| `check_synapse_id` | Validate a Synapse ID exists | `operations.is_synapse_id()` |
| `md5_query` | Find entities by file MD5 hash | `operations.md5_query()` |
| `get_docker_repository` | Get DockerRepository metadata | `DockerRepository(id).get()` |

## Return type wrapping

MCP requires dict responses. SDK primitives are wrapped:

```mermaid
flowchart LR
    A["is_synapse_id() → bool"] --> B["{syn_id, is_valid: true}"]
    C["md5_query() → list"] --> D["{md5, results: [...]}"]
    E["find_entity_id() → str"] --> F["{entity_id, name, parent_id}"]
    G["DockerRepository.get() → dataclass"] --> H["serialize_model → dict"]
```

## Complete epic tool count

With this PR merged, the MCP server exposes **51 tools** across 15 domains:

| Domain | Tools |
|--------|-------|
| Entity Core | 5 |
| Entity ACL | 3 |
| Entity Schema | 4 |
| Container Traversal | 1 |
| Activity | 2 |
| Link | 1 |
| Wiki | 4 |
| Team & User | 6 |
| Evaluation | 4 |
| Submission | 8 |
| Schema Organizations | 6 |
| Curation Tasks | 3 |
| Forms | 1 |
| Utility (4.12.0) | 3 |
| Docker | 1 |

## Test plan

- [x] All 125 tests pass
- [x] All 51 tools verified against live Synapse in prior testing

[DPE-1621]: https://sagebionetworks.jira.com/browse/DPE-1621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #37, opened 2026-04-24
[DPE-1641] BM25 tool discovery: @service_tool decorator, synonyms, eval harness

BryanFauble:
## Summary

Stacks on top of [#29](https://github.com/Sage-Bionetworks/synapse-mcp/pull/29). With the DPE-1622 → DPE-1628 stack the Synapse MCP catalog grew from 8 tools to **50**. At that size, LLM tool-selection accuracy degrades unless names, descriptions, synonyms, and disambiguation are deliberately engineered and the client no longer loads the full catalog on every call. This PR adds the discovery infrastructure and backfills the metadata.

Jira: [DPE-1641](https://sagebionetworks.jira.com/browse/DPE-1641)

## What changes

- New `@service_tool` decorator (`src/synapse_mcp/services/tool_service.py`) — the single registration path for every tool going forward. Validates naming, validates the first sentence of the description names the concrete Synapse object, appends a `Related terms:` line (synonyms for BM25 recall) and a `Distinct from:` line (siblings for LLM disambiguation), applies operation-driven MCP annotations, and wraps the tool with exception handling.
- All 50 tools in `src/synapse_mcp/tools.py` converted. User-intent-led first sentences, ID-example values on every ID parameter, synonyms reflecting real user language.
- Three non-standard names renamed to match the approved verb-prefix set:
  - `find_entity_id` → `search_entity_by_name`
  - `is_user_certified` → `check_user_certified`
  - `md5_query` → `search_entities_by_md5`
- `BM25SearchTransform` installed on the FastMCP server. `get_entity` and `search_synapse` remain always-visible; everything else is reached via the synthetic `search_tools` tool.
- `doc/tool-authoring.md` documents the convention.
- 21 new decorator unit tests and a 57-query BM25 eval harness under `tests/evals/`.

## Why BM25 tool discovery

Current (pre-PR) behaviour: every MCP client call to `list_tools` returns all 50 tool definitions, ~1000+ tokens of catalog context the LLM must hold before selecting one. That scales poorly, and with heavy semantic overlap (entity/activity/submission, wiki/docs, evaluation/challenge, ACL vs permissions vs sharing, Docker repository vs container runtime) the LLM frequently picks the wrong neighbour.

New behaviour: the LLM sees a 4-tool catalog (`get_entity`, `search_synapse`, `search_tools`, `call_tool`) by default. When it needs a capability outside those two, it calls `search_tools` with a natural-language query and gets back the top-7 BM25-ranked matches. BM25 runs against the combined text of each tool's name, description, parameter names, and parameter descriptions.

### Tool discovery flow

```mermaid
sequenceDiagram
    participant LLM
    participant FastMCP
    participant BM25 as BM25SearchTransform
    participant ToolReg as ServiceTool catalog

    LLM->>FastMCP: list_tools
    FastMCP->>ToolReg: raw catalog (50 tools)
    ToolReg->>BM25: full catalog
    BM25-->>FastMCP: always_visible + search_tools + call_tool
    FastMCP-->>LLM: 4-tool catalog

    LLM->>FastMCP: search_tools(query)
    FastMCP->>BM25: invoke synthetic search tool
    BM25->>BM25: BM25 rank against full catalog
    BM25-->>LLM: top-7 ranked tools

    LLM->>FastMCP: call_tool(tool_name, args)
    FastMCP->>ToolReg: invoke wrapped fn
    ToolReg-->>LLM: result
```

### Why the @service_tool decorator

Rather than letting each tool author hand-roll `@mcp.tool(...)` calls, every tool goes through one decorator that enforces the conventions. This moves BM25 engineering from a "remember to do it" habit to a registration-time hard gate.

```mermaid
flowchart LR
    A["service_tool decorator called"] --> B{prefix in approved set?}
    B -- no --> Z1[ValueError at import]
    B -- yes --> C{first sentence names synapse_object?}
    C -- no --> Z2[ValueError at import]
    C -- yes --> D[append Related terms synonyms]
    D --> E[append Distinct from siblings]
    E --> F[pick MCP annotations by operation]
    F --> G[wrap fn with exception handler]
    G --> H[register via mcp.tool]
```

## BM25-friendly description anatomy

```mermaid
flowchart TB
    P[Primary description]
    S[Related terms line]
    D[Distinct from line]

    P -->|LLM reads first| L1[Use this when the user wants to know what produced a Synapse entity]
    S -->|BM25-indexed aliases| L2[lineage, history, inputs, outputs, derived from, provenance record]
    D -->|LLM uses for disambiguation| L3[get_activity, get_entity]

    style P fill:#e1f5ff
    style S fill:#fff4e1
    style D fill:#ffe1e1
```

Primary description is written for a human; synonyms drive BM25 recall when the user says "lineage" instead of "provenance"; siblings let the LLM disambiguate close neighbours.

### Tool-selection eval baseline

57 natural-language queries spanning all 14 services, using synonyms and user language (not Synapse jargon). BM25 runs directly against the raw catalog (`tests/evals/test_tool_selection.py`):

| Metric | Result | Gate |
|--------|--------|------|
| top-1 accuracy | **86.0%** | ≥ 70% ✅ |
| top-3 accuracy | **100.0%** | ≥ 90% ✅ |
| top-5 accuracy | **100.0%** | — |

Sample queries and the tool BM25 ranks #1:

| Query (user language) | Ranked tool |
|------------------------|-------------|
| "what produced syn456" | `get_entity_provenance` |
| "who is on the NF team" | `get_team_members` |
| "show me the readme for this project syn100" | `get_wiki_page` |
| "is user 1234567 certified" | `check_user_certified` |
| "find all files with cancer_type=glioma" | `search_synapse` |
| "what's the Synapse Docker repo for syn789" | `get_docker_repository` |
| "does syn999 exist" | `check_synapse_id` |
| "sharing settings on syn333" | `get_entity_acl` |

### Catalog surface reduction

| | Before | After |
|--|--------|-------|
| Tools returned by `list_tools()` | 50 | 4 (`get_entity`, `search_synapse`, `search_tools`, `call_tool`) |
| Discovery context tokens | ~1000+ | ~150 |

Every tool is still callable — the LLM reaches them by calling `search_tools(query=...)` and then `call_tool(tool_name, args)`.

## Key decisions and their rationale

1. **Decorator enforced at registration time, not at review time.** The first sentence must name the Synapse object (validated against `synapse_object=`), and the name must start with an approved verb. A typo makes the whole module fail to import, which is loud enough to catch before merging.
2. **Description kwarg, not docstring, is the source of truth.** Existing tools already pass `description=` to `@mcp.tool`, so we continue to treat that as the LLM-visible copy. Docstrings are kept as short one-liners for Python devs.
3. **`Related terms:` as a separate line, not embedded in prose.** Keeps the primary description readable while still feeding the BM25 tokenizer. The tokenizer (`re.split(r"[^a-z0-9]+")`, filter `len > 1`) doesn't care about paragraph structure; the human reader does.
4. **`Distinct from:` uses sibling tool names verbatim.** BM25 tokenizes the sibling names, which are already word-rich (`get_entity_provenance` → `{get, entity, provenance}`). Listing them also trains the LLM to pick the right one when it sees multiple candidates.
5. **`always_visible = ["search_synapse", "get_entity"]`.** These two cover the most common first step for any Synapse workflow — look up a known ID, or find entities by keyword — so the LLM never has to call `search_tools` to discover them. Every other capability goes through `search_tools` / `call_tool`.
6. **Double error handling (service + tool wrapper) is intentional.** Service methods already use `@error_boundary`. The tool wrapper's try/except catches anything thrown BEFORE the service call — e.g. in `validate_synapse_id` — which previously escaped uncaught.
7. **`operation` kwarg drives MCP annotations only.** It tags the tool for observability (`readonly`, `mutation`, `destructive`, `admin` tags) and picks the right MCP annotation block (`readOnlyHint`, `destructiveHint`). It does NOT gate visibility.

## Out of scope

- No changes to the existing stacked branches (DPE-1622 → DPE-1628).
- No new write / destructive / admin tools. Catalog stays 100% read-only.
- PR #10 (curation hackathon branch) is not rebased; it predates the service layer and needs its own session.
- No CI workflow changes. The eval runs under the existing `uv run pytest` invocation and prints top-1/top-3/top-5 accuracy to the test output.
- No tool visibility gating. The decorator still tags tools informationally but nothing currently consumes those tags for visibility filtering.

## Test plan

- [x] `uv run pytest` — 220 tests pass (162 prior + 21 new decorator tests + 37 eval harness tests).
- [x] `uv run pytest tests/evals/test_tool_selection.py -s` — eval prints top-1/3/5 summary, all 57 queries in top-3.
- [x] Start the server (`uv run synapse-mcp`) with the default env and call `tools/list`: returns 4 tools (`get_entity`, `search_synapse`, `search_tools`, `call_tool`).
- [x] Call `search_tools(query="...")` and confirm BM25 returns sensible rankings.

## Review guidance

For BM25-specific justification inside the code itself, see the review thread attached to this PR on the Files Changed tab. Key review hotspots:
- `services/tool_service.py` — the decorator, synapse-object validation, synonym/sibling appending
- `tools.py` — a representative sample of the 50 converted tools, especially clusters with high token overlap (submissions, wiki, schema)
- `tests/evals/test_tool_selection.py` — the query fixtures; feedback welcome on whether any realistic user phrasing is missing

[DPE-1641]: https://sagebionetworks.jira.com/browse/DPE-1641?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/synapse_mcp/services/tool_service.py
```diff
@@ -206,3 +217,260 @@ async def wrapper(self, ctx, *args, **kwargs):
...
+
+def _first_sentence(text: str) -> str:
+    """Return the first sentence of ``text`` (up to the first period+space)."""
+    for sep in (". ", ".\n", ".\t"):
+        if sep in text:
+            return text.split(sep, 1)[0].strip() + "."
+    return text.strip()
+
+
+def _synapse_object_forms(obj: str) -> Tuple[str, ...]:
```
BryanFauble, line 294: **Why match on the head noun, not the full phrase?**

The strict `"synapse entity" in first_sentence` check I started with broke on natural copy — e.g. `synapse_object="Synapse entity"` wouldn't match a first sentence that says `"every Synapse entities"` (plural) or `"a Synapse Link entity"` (subtype prefix). Neither of those is a BM25 problem — they're still clearly about entities — but the decorator was rejecting the registration.

Head-noun matching (`entity` / `entities`) picks up both forms while still catching the thing we actually care about: a typo'd or absent mention of what kind of Synapse object the tool operates on. For `"Synapse JSON Schema"` the head noun is `"schema"`, which matches `"schema"`, `"schemas"`, `"JSON schema"`, etc. For `"Synapse Docker repository"` it's `"repository"` / `"repositories"`.

BM25 ranking itself is untouched — this is just registration-time input validation.

src/synapse_mcp/services/tool_service.py
```diff
@@ -206,3 +217,260 @@ async def wrapper(self, ctx, *args, **kwargs):
...
+    if operation in ("write", "destructive"):
+        tags.add("mutation")
+    if operation == "destructive":
+        tags.add("destructive")
+    if operation == "admin":
+        tags.add("admin")
+    return frozenset(tags)
+
+
+def _build_extended_description(
```
BryanFauble, line 337: **Why two separate lines, not inline prose?**

BM25's tokenizer (`re.split(r"[^a-z0-9]+", text.lower())`, filter `len > 1`) treats `"Related terms: lineage, history"` identically to `"...including lineage and history..."` — both contribute the same tokens to the index. But the *human reader* — both the LLM picking a tool and any engineer debugging — benefits from the primary description being short and prose-like, with the alias list clearly labelled and skippable.

Two effects from keeping the sections separate:

1. The primary description stays below the \~200-token sweet spot for LLM reading. Burying synonyms in prose pushes the description past that threshold and actually hurts selection, not helps.
2. When the LLM sees `Distinct from: get_wiki_history, get_wiki_page`, it reads it as disambiguation metadata. When the same names are embedded in a sentence, it has to reason about why they're there.

This is the single most load-bearing piece of BM25 plumbing in the PR. Everything else is cleanup.

src/synapse_mcp/services/tool_service.py
```diff
@@ -206,3 +217,260 @@ async def wrapper(self, ctx, *args, **kwargs):
...
+    }
+    response = getattr(exc, "response", None)
+    if response is not None:
+        status_code = getattr(response, "status_code", None)
+        if status_code is not None:
+            err["status_code"] = status_code
+    return err
+
+
+def service_tool(
```
BryanFauble, line 388: **Why validate at registration time, not review time?**

The Phase 1 audit (pre-PR) found 42/50 tool descriptions led with API mechanics (`"Returns the..."`, `"Fetches the..."`) instead of user intent, and 44/50 had ID parameters with no example value. Those are the kinds of regressions a reviewer misses on tool #47 of a dependency PR.

Moving the checks into the decorator means the module fails to import the moment someone writes `@service_tool(...)` with a non-approved prefix or a description that doesn't name its Synapse object. You can't ship a regression past this unless you deliberately skip the decorator — and the convention doc explicitly says not to.

The cost is that adding a new approved prefix (`validate_`, `register_`, etc.) is a code change rather than a review-time note. That trade is acceptable because the prefix set is small and stable.

src/synapse_mcp/tools.py
```diff
@@ -19,49 +29,98 @@
...
-    "idempotentHint": True,
-    "destructiveHint": False,
-    "openWorldHint": True,
-}
+
+# Reusable synonym sets so BM25 indexes user-language aliases for every
+# relevant tool without copy-pasting the same list 8 times. Keep these
+# tight: only include aliases users actually say, not every tangential
+# synonym.
+_ENTITY_TYPES = (
```
BryanFauble, line 40: **Why the shared synonym constants?**

Eleven tools all operate on a Synapse entity that could be a project, folder, file, table, view, or dataset. Without `_ENTITY_TYPES` I'd be copying the same 7-element tuple into every decoration, and the next time someone adds `dataset_collection_link` or renames `dataset_collection` we'd have eleven places to update and at least one would get forgotten.

The same logic drives `_EVALUATION_SYNONYMS`, `_SUBMISSION_SYNONYMS`, `_PROVENANCE_SYNONYMS`, etc. — each one is a stable vocabulary shared across a cluster of tools. Per-tool synonyms that *aren't* shared (e.g. `"sharing settings"` on `get_entity_acl`) stay inline at the call site.

Keep these lists tight. They exist for BM25 recall — if you add a synonym that isn't what real users say, you're just adding noise to the index and pushing the relevant signal down.

src/synapse_mcp/tools.py
```diff
@@ -654,15 +992,32 @@ async def get_submission(
...
+    + _EVALUATION_SYNONYMS
+    + ("all entries", "all submissions", "everyone"),
+    siblings=(
+        "list_submission_statuses",
+        "list_evaluation_submission_bundles",
+        "list_my_submissions",
     ),
-    annotations=_RO,
 )
 async def list_evaluation_submissions(
```
BryanFauble, line 1034: **Why `list_my_submissions` is in the siblings list (and spelled out in the description)**

The audit flagged `list_evaluation_submissions` ↔ `list_submission_statuses` as the only Jaccard ≥ 0.7 cluster in the whole catalog — both mention "list", "submission", "evaluation", "status", and several function words. Without disambiguation BM25 will rank them indistinguishably for queries like "submissions to challenge 9600001".

The fix has two parts:
1. **Primary description explicitly contrasts** — "NOT just the caller's own — use `list_my_submissions` for that." Puts the disambiguation in prose the LLM reads first.
2. **`siblings=` names the three nearest neighbours** (`list_submission_statuses`, `list_evaluation_submission_bundles`, `list_my_submissions`). The `Distinct from:` line rendered from this feeds BM25 the sibling names as additional index tokens, so when the LLM sees `list_my_submissions` in the ranked output it can pick correctly.

After this treatment the eval query `"all entries to challenge 9600001"` went from rank 7 to rank 1. This is the pattern to repeat any time the audit flags a cluster.

src/synapse_mcp/tools.py
```diff
@@ -1045,14 +1606,30 @@ async def md5_query(
...
+    synonyms=(
+        "synapse docker",
+        "docker image",
+        "image repository",
+        "container registry entity",
     ),
-    annotations=_RO,
+    siblings=("get_entity", "get_entity_acl"),
 )
 async def get_docker_repository(
```
BryanFauble: **The `NOT container runtime` callout is deliberate.**

"Docker" is the single most overloaded word in this catalog. Users asking a general LLM about Docker mean Docker the daemon, Docker the CLI, a running container, a registry push, a Compose file — almost never a Synapse entity that *references* a Docker image stored in Synapse's registry.

If this tool's description said `"Get a Synapse Docker repository entity by ID"` and stopped there, any query that looks vaguely Docker-shaped (`"start the docker container"`, `"pull my docker image"`, `"docker compose up"`) would BM25-rank this tool highly because it has a near-monopoly on the `docker` token in the catalog.

The explicit negation gives the LLM a signal when it reads the description: if the user's intent is about container runtimes rather than Synapse metadata, this is the wrong tool. The synonyms (`"synapse docker"`, `"image repository"`, `"container registry entity"`) further anchor it to the Synapse context.

This is the one tool in the PR where I'd recommend future authors **keep the exclusion language** even if it feels awkward.

src/synapse_mcp/tools.py
```diff
@@ -1063,3 +1640,26 @@ async def get_docker_repository(
...
+# Discovery: BM25 search transform
+# ---------------------------------------------------------------------------
+# Applied after all tools are registered so the transform has the full
+# catalog to index. The LLM's default view becomes the two pinned tools
+# plus the synthetic ``search_tools`` / ``call_tool`` pair; every other
+# tool is reached by calling ``search_tools`` with a natural-language
+# query and then invoking ``call_tool``.
+
+
+def _configure_discovery_transforms() -> None:
```
BryanFauble, line 1654: **Why `always_visible = ["search_synapse", "get_entity"]`?**

Every Synapse workflow has one of two entry points:

- The user already knows the Synapse ID and wants details → `get_entity`.
- The user wants to find an entity by keyword/topic → `search_synapse`.

Pinning these two means the LLM can handle the common 80% of first-turn interactions without routing through `search_tools` first. Any other capability (provenance, wiki, teams, evaluations, submissions, schemas, Docker, curation, forms, utility) is one extra step — the LLM calls `search_tools(query=...)` and the BM25 index surfaces the right tool in top-3 (currently 100% top-3 on the 57-query eval).

**Why `max_results=7`?** BM25 returns 7 ranked candidates per search. Three is too tight for a close cluster (submissions has 8 siblings); ten starts eating LLM context budget for low-relevance matches. Seven lets the LLM see the near-miss alternatives without drowning in them.

**Why no `Visibility` transform?** Out of scope for this PR — the tags are already computed by `@service_tool` so the hook is there for whenever we add destructive/admin operations, but we have none today.

tests/evals/test_tool_selection.py
```diff
@@ -0,0 +1,182 @@
...
+    ("does syn999 exist", "check_synapse_id"),
+    ("find the file with md5 9e107d9d372bb6826bd81d3542a419d6", "search_entities_by_md5"),
+    ("what's the synapse id of the file named sample.csv in folder syn100", "search_entity_by_name"),
+    # docker
+    ("what's the Synapse Docker repo for syn789", "get_docker_repository"),
+    ("docker image entity by synapse id syn789", "get_docker_repository"),
+]
+
+
+async def _build_index() -> Tuple[_BM25Index, list]:
```
BryanFauble, line 130: **Why `_list_tools()` (private) and not `list_tools()` (public)?**

This tripped me up during iteration. `mcp.list_tools()` runs *through* the registered transforms — so by the time we see the output, `BM25SearchTransform` has already replaced the catalog with `[search_tools, call_tool, get_entity, search_synapse]`. Running the eval against that 4-tool list defeats the point.

`mcp._list_tools()` returns the raw 50-tool catalog before any transform runs. This is what the BM25 transform *itself* indexes internally at runtime (see `BM25SearchTransform._search()` in the fastmcp source), so reproducing the same index in the eval is the most faithful test of production ranking behaviour.

If `_list_tools` ever becomes public or the wrapping changes, this is the line to update. Leaving the underscore-prefix usage visible (rather than wrapping it) so grep hits it immediately.

BryanFauble (commented):
Design-rationale comments for the BM25 + discovery decisions in this PR. These are explanations of *why* the code looks the way it does, not requests for changes — feel free to resolve each thread once you've read it.

Hotspots called out below:
- `tool_service.py`: head-noun matching, `Related terms:` / `Distinct from:` line separation, registration-time validation
- `tools.py`: shared synonym constants, Jaccard-cluster disambiguation, Docker-repository exclusion language, always-visible choices
- `test_tool_selection.py`: why the eval uses the private `_list_tools()`

src/synapse_mcp/services/tool_service.py
```diff
@@ -235,3 +246,263 @@ async def wrapper(self, ctx, *args, **kwargs):
...
+            follow the pattern "Use this when <user intent>.
+            <Synapse object> is <one-line definition>." and must name
+            ``synapse_object``.
+        synonyms: Common user-language aliases for the Synapse terms
+            in this tool (e.g. ``("lineage", "history")`` for
+            provenance). Appended to the description so the BM25
+            index matches them, without polluting the primary copy.
+        siblings: Names of closely-related tools. Appended as a
+            "Distinct from:" line so the LLM sees the disambiguation.
+        annotations: Override for MCP tool annotations. When omitted,
```
linglp, line 424: No tool in `tools.py` passes `annotations= explicitly`. Every tool relies entirely on `_default_annotations_for(operation)`. I am curious how we could use it in the future? It seems like that the options are: `readOnlyHint`, `idempotentHint`, `destructiveHint`, and `openWorldHint`. Is that true? If so, can we add that to the documentation?

linglp (commented):
Thanks @BryanFauble ! It looks good overall. I tried: 

1. `help me find syn66077177`
Ther response: 
````
syn66077177 is a Project named test-json-schema-linglp.

Key details:

Name: test-json-schema-linglp
Parent: syn4489
Created: 2025-04-10
Annotations: none
Contents: empty (no files, folders, tables, or other entities)
````
Based on the raw return from `operations_get_async`, the response makes sense. But this is also confusing because this project is not empty. Later, when I tried to search for entities that contain recordset or folder that contains subfolders, it seems like that "contents" is always empty. 

2. `what's the difference between version 1 and version 6 of syn74767777`
Here's the response: `The available tools don't include a "get entity by version" lookup, so I can't directly compare the content of v1 vs v6. Let me check the provenance for both versions to see if there's lineage info`. 

Is this expected? I think in theory, we should be able to do something like: `get_async(synapse_id="syn123456", version_number=2)`

3. `is there docker container stored in project syn66077177`? 
Here's the response: `No Docker repositories were found directly in project syn66077177. The search returned 0 hits for dockerrepo entity type under that parent.It's possible Docker repos could be nested deeper in subfolders rather than directly in the project root. Would you like me to search recursively through the project's children?`
This is inaccurate as well. There's a docker repo there. 

I will test more but I just want to surface these problems that I encountered first.

File: src/synapse_mcp/services/tool_service.py (hunk 9 above)
Comment (BryanFauble) [reply src/synapse_mcp/services/tool_service.py:424]: Right, today every tool relies on `_default_annotations_for(operation)` and the four MCP hint keys you listed (`readOnlyHint`, `idempotentHint`, `destructiveHint`, `openWorldHint`) are the full set the spec defines. There are three default sets keyed off `operation`:

- `read` → `DEFAULT_READ_ANNOTATIONS` (read-only, idempotent)
- `write` → `DEFAULT_WRITE_ANNOTATIONS` (mutating, non-idempotent, non-destructive)
- `destructive` → `DEFAULT_DESTRUCTIVE_ANNOTATIONS` (mutating, non-idempotent, destructive)

The future use case is when an individual tool needs to override one of those hints — e.g. a write that's actually idempotent (set-by-id-style upsert), or an `openWorldHint=False` tool that only touches a fixed local resource. The decorator already accepts an explicit `annotations=` kwarg that wins over the operation default, so it's a per-tool override knob rather than something we'd document for the typical case.

I'll add a short note to `doc/tool-authoring.md` calling out the four keys and when to override.

File: src/synapse_mcp/services/tool_service.py [hunk 10]
```diff
@@ -192,11 +209,272 @@ async def wrapper(ctx, *args, **kwargs):
...
+    Includes alias tags (``readonly``, ``mutation``, ``destructive``,
+    ``admin``) so a single Visibility rule can target a broad category
+    without listing every specific service name.
+    """
+    tags = {service, operation}
+    if operation == "read":
+        tags.add("readonly")
+    if operation in ("write", "destructive"):
+        tags.add("mutation")
+    if operation == "destructive":
```
linglp: Nit: we could tighten this up a little bit: 
```
    if operation in ("destructive", "admin"):
        tags.add(operation)
```
And you don't need: 
```
    if operation == "admin":
        tags.add("admin")
```

src/synapse_mcp/services/tool_service.py
```diff
@@ -192,11 +209,272 @@ async def wrapper(ctx, *args, **kwargs):
...
+
+    def decorator(fn: Callable) -> Callable:
+        if not any(fn.__name__.startswith(p) for p in VALID_TOOL_PREFIXES):
+            raise ValueError(
+                f"Tool '{fn.__name__}' must start with one of "
+                f"{VALID_TOOL_PREFIXES}"
+            )
+
+        first = _first_sentence(description).lower()
+        forms = _synapse_object_forms(synapse_object)
```
linglp, line 441: Note: This is to ensure the tool’s description first sentence includes the declared object. For example: 
```
from synapse_mcp.services.tool_service import _synapse_object_forms
synapse_objects = "Synapse FormGroup"

result = _synapse_object_forms(synapse_objects)
print(result)
```
this prints `'formgroups', 'synapse formgroup', 'formgroup'`, and in the description, it has to mention one of the three.

src/synapse_mcp/services/wiki_service.py
```diff
@@ -36,39 +33,33 @@ async def get_wiki_page(
...
+                async for header in WikiHeader.get_async(
+                    owner_id=owner_id,
+                    offset=0,
+                    limit=50,
+                    synapse_client=client,
+                ):
+                    if getattr(header, "parent_id", None) is None:
+                        root = header
+                        break
+                if root is None:
```
linglp, line 55: Nit: I tried to find the test for this part, but I don't think this part is directly tested?

src/synapse_mcp/services/wiki_service.py
```diff
@@ -171,35 +145,26 @@ async def get_wiki_history(
...
-            owner_id: The Synapse ID of the wiki owner.
+            owner_id: Synapse ID of the wiki owner.
 
         Returns:
-            Dict with owner_id, owner_object_type, id_list (page ordering),
+            Dict with owner_id, id_list (page ordering),
             and etag.
-            Or if the wiki order hint is not set, returns an error.
         """
-        if not owner_id:
```
linglp, line 193: Nit: Could we add/point to a tool-layer test proving missing `owner_id` is rejected before `WikiService.get_wiki_order_hint` is invoked? I checked `TestServiceToolErrorBoundary`, but I didn't find the test.

linglp (approved):
LGTM! Just some nit comments!

PR #39, opened 2026-06-18
[DPE-1728] Address security vulnerabilities 

linglp:
# **Problem:**
Address security vulnerabilities: 
Screenshot 2026-06-22 at 12 31 31 PM

33 Dependabot security alerts across 10 packages: starlette, python-multipart, authlib, urllib3, idna, pygments, pyjwt, cryptography, requests, and protobuf. Additionally, uv.lock was out of sync — it had synapseclient 4.10.0 locked despite pyproject.toml pinning ==4.12.0, which was keeping urllib3 stuck at 1.26.x.

# Changes:
* Added starlette>=1.3.1 as a direct dependency — it is imported directly in app.py and auth_middleware.py 
* Regenerated uv.lock
* Raised lower bounds on directly declared deps 

# Notes
I was thinking of adding security floor constraints under `[tool.uv].constraint-dependencies` for purely transitive deps (python-multipart, authlib, urllib3, idna, protobuf, pygments). But the constraints won't work if we install the package like the documentation `pip install -e .`


BryanFauble (approved):
LGTM!

PR #45, opened 2026-07-07
Generate README tool table from the tool catalog

BryanFauble:
# **Problem:**

The README's **Available Tools** section documented only **5 tools** (`get_entity`, `get_entity_annotations`, `get_entity_provenance`, `get_entity_children`, `search_synapse`), but `src/synapse_mcp/tools.py` now registers **48 read-only tools across 12 service domains** (entity, search, activity, schema, wiki, team, user, evaluation, submission, curation, organization, form, utility). The "You can" capability list and the Tool Discovery paragraph (which said "50-tool catalog") were likewise stale. A hand-maintained table of 48 tools is guaranteed to drift as tools are added or reworded.

# **Solution:**

Make the table **generated from the live catalog** and **kept in sync automatically**, so it can't drift. Only the table is generated; surrounding prose stays hand-written so the doc reads well.

- **`scripts/gen_tool_table.py`** — introspects `mcp._list_tools()` (the pre-transform catalog, the same path the BM25 eval uses) and renders `| Tool | Domain | Description |` between `BEGIN/END GENERATED TOOLS` markers. Tool column shows required params only (`get_entity(entity_id)`, `search_synapse()`); Description is the first sentence (with an abbreviation guard so `e.g.`/`i.e.` don't truncate); rows follow a curated domain order mirroring `tools.py`, alphabetical within a domain. Unknown domains append with a warning (never silently dropped). Modes: `--write` (auto-fix, non-zero if it changed the file) and `--check` (never writes, prints a unified diff on drift).
- **`README.md`** — expanded the capability summary, inserted the generated table, added a "all tools are read-only" note, and rewrote the Tool Discovery paragraph to describe the actual mechanism (always-visible `get_entity`/`search_synapse` + synthetic `search_tools`/`call_tool`).
- **`.pre-commit-config.yaml`** — first pre-commit config in the repo; a `repo: local`, `language: system` `sync-readme-tools` hook that auto-fixes the table when `tools.py`, the script, or `README.md` change.
- **`tests/test_readme_tools.py`** — subprocesses `--check`, so the existing CI `pytest` job fails (with the diff) on drift — no new CI job.
- **Dev deps / docs** — added `pre-commit` to `requirements-dev.txt` and the `pyproject.toml` dev group; documented `pre-commit install` in `DEVELOPMENT.md`.

**Design decisions:** three enforcement layers (script + pre-commit + CI test) so drift is caught before push *and* at merge; pre-commit auto-fixes locally while CI is check-only (must not mutate the tree); the generator's `build_table` logic is shared via the single `--check` path.

Note: `uv.lock` updated because `pre-commit` was added to the dev group. A couple of pre-existing "50 tools" comments in `tests/` were left out of scope — the generated table is now the single source of truth for the count.

# **Testing:**

- `uv run python scripts/gen_tool_table.py` produces a 48-row table matching `grep -c '^async def ' src/synapse_mcp/tools.py`; second run is a no-op (idempotent).
- `--check` exits 0 when in sync; corrupting a row makes it exit 1 and print a unified diff; reverting restores exit 0.
- Pre-commit hook: passes on a clean tree; on a stale README it reports "files were modified" and rewrites the table, then passes on re-run (byte-identical to correct output).
- `uv run pytest tests/test_readme_tools.py` passes; full suite: **307 passed**.

### DIFF
--- a/scripts/gen_tool_table.py
+++ b/scripts/gen_tool_table.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python3
+"""Generate the Available Tools table in README.md from the live catalog.
+
+Introspects the registered FastMCP tools (the same pre-transform catalog the
+BM25 eval consumes) and renders a Markdown table that is injected into
+README.md between the BEGIN/END markers. Keeping the table generated means it
+never drifts from ``src/synapse_mcp/tools.py``.
+
+Usage:
+    uv run python scripts/gen_tool_table.py            # rewrite README.md
+    uv run python scripts/gen_tool_table.py --write     # same as default
+    uv run python scripts/gen_tool_table.py --check      # verify, never write
+
+``--write`` exits non-zero if it changed the file (pre-commit convention).
+``--check`` exits non-zero if the table is stale, printing a unified diff.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import difflib
+import os
+import sys
+from pathlib import Path
+from typing import Iterable
+
+# app.py raises at import time without auth configured; set a dummy PAT before
+# importing synapse_mcp (same trick as tests/conftest.py).
+os.environ.setdefault("SYNAPSE_PAT", "readme-gen")
+
+from synapse_mcp import mcp  # noqa: E402
+from synapse_mcp.services.tool_service import ServiceName  # noqa: E402
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+README_PATH = REPO_ROOT / "README.md"
+
+BEGIN_MARKER = (
+    "<!-- BEGIN GENERATED TOOLS: run "
+    "`uv run python scripts/gen_tool_table.py` to update -->"
+)
+END_MARKER = "<!-- END GENERATED TOOLS -->"
+
+# Curated domain order, mirroring the "Domain N" sections in tools.py so the
+# README and source tell the same story. Tools with a domain not listed here
+# are appended alphabetically after a warning (never silently dropped).
+DOMAIN_ORDER = [
+    "entity",
+    "search",
+    "activity",
+    "schema",
+    "wiki",
+    "team",
+    "user",
+    "evaluation",
+    "submission",
+    "curation",
+    "organization",
+    "form",
+    "utility",
+]
+
+# Service-name tags, used to pick the domain tag out of the full tag set
+# (which also carries alias tags like "read"/"readonly").
+_SERVICE_TAGS = set(ServiceName.__args__)
+
+
+def _domain_for(tool) -> str:
+    """Return the single service-domain tag for a tool."""
+    domains = sorted(set(tool.tags) & _SERVICE_TAGS)
+    if not domains:
+        return "?"
+    return domains[0]
+
+
+def _signature(tool) -> str:
+    """Render ``name(required_param, ...)`` from the tool's parameter schema.
+
+    ``ctx`` is already excluded by FastMCP. Only required params are shown;
+    tools whose args are all optional render as ``name()``.
+    """
+    params = getattr(tool, "parameters", None) or {}
+    required = params.get("required", []) or []
+    return f"{tool.name}({', '.join(required)})"
+
+
+# Abbreviations whose trailing period is not a sentence end. Compared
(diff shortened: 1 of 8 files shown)


scripts/gen_tool_table.py
```diff
@@ -0,0 +1,230 @@
...
+Usage:
+    uv run python scripts/gen_tool_table.py            # rewrite README.md
+    uv run python scripts/gen_tool_table.py --write     # same as default
+    uv run python scripts/gen_tool_table.py --check      # verify, never write
+
+``--write`` exits non-zero if it changed the file (pre-commit convention).
+``--check`` exits non-zero if the table is stale, printing a unified diff.
+"""
+
+from __future__ import annotations
```
thomasyu888: How are we benefitting from this in this script?

scripts/gen_tool_table.py
```diff
@@ -0,0 +1,230 @@
...
+import asyncio
+import difflib
+import os
+import sys
+from pathlib import Path
+from typing import Iterable
+
+# app.py raises at import time without auth configured; set a dummy PAT before
+# importing synapse_mcp (same trick as tests/conftest.py).
+os.environ.setdefault("SYNAPSE_PAT", "readme-gen")
```
thomasyu888, line 28: I didn't notice at the time, and was there a reason we went with `SYNAPSE_PAT` over `SYNAPSE_AUTH_TOKEN`?

thomasyu888 (commented):
🔥 This is a neat way to keep the readme table up to date using the text anchors, and I'll defer to the team for final review.

scripts/gen_tool_table.py
```diff
@@ -0,0 +1,230 @@
...
+
+def _domain_for(tool) -> str:
+    """Return the single service-domain tag for a tool."""
+    domains = sorted(set(tool.tags) & _SERVICE_TAGS)
+    if not domains:
+        return "?"
+    return domains[0]
+
+
+def _signature(tool) -> str:
```
andrewelamb: Could we add typing to this and the other functions:

```
from fastmcp.tools import Tool   # the base class, not FunctionTool

def _signature(tool: Tool) -> str:
```

scripts/gen_tool_table.py
```diff
@@ -0,0 +1,230 @@
...
+# (which also carries alias tags like "read"/"readonly").
+_SERVICE_TAGS = set(ServiceName.__args__)
+
+
+def _domain_for(tool) -> str:
+    """Return the single service-domain tag for a tool."""
+    domains = sorted(set(tool.tags) & _SERVICE_TAGS)
+    if not domains:
+        return "?"
+    return domains[0]
```
andrewelamb, line 72: Will a tool only ever have one service-domain?

scripts/gen_tool_table.py
```diff
@@ -0,0 +1,230 @@
...
+    "utility",
+]
+
+# Service-name tags, used to pick the domain tag out of the full tag set
+# (which also carries alias tags like "read"/"readonly").
+_SERVICE_TAGS = set(ServiceName.__args__)
+
+
+def _domain_for(tool) -> str:
+    """Return the single service-domain tag for a tool."""
```
andrewelamb, line 68: It seems like this gets run on tools in multiple locations. Should we run this iteratively over oevery tool once at the top of the main function nad pass the reuslts to the functiosn that need it?

andrewelamb (approved):
LGTM! I just had a few questions.

scripts/gen_tool_table.py
BryanFauble: Yes there is only a single domain

scripts/gen_tool_table.py
BryanFauble: Patched

scripts/gen_tool_table.py
BryanFauble: Swapped to run once

scripts/gen_tool_table.py
BryanFauble: This is what was in the original github repo that we inherited and I never changed it.

scripts/gen_tool_table.py
BryanFauble: Removed it

PR #47, opened 2026-07-08
[DPE-1757] Add write/destructive (CRUD) tools to the MCP server

BryanFauble:
# **Problem:**

The Synapse MCP server was read-only — every tool used `operation="read"`. There was no way for an LLM to create, update, or delete Synapse objects through the server, and no mechanism to let a client permit reads while blocking writes.

Relevant code: `src/synapse_mcp/tools.py` (tool registrations), `src/synapse_mcp/services/*` (SDK wrappers), and the discovery transform registered at the bottom of `tools.py`.

# **Solution:**

Adds the **file-free** write surface (create/update/delete) the Synapse Python client supports — no file content is ever uploaded or downloaded.

**24 new write/destructive tools** across the entity, schema, team, evaluation, submission, organization, and curation domains (see the regenerated README tool table for the full list). They reuse the existing `@service_tool` machinery: `operation="write"`/`"destructive"`, thin tool shims delegating to `@error_boundary` service methods.

**Split dispatch proxy** — a new `SplitCallTransform` (`src/synapse_mcp/discovery.py`, subclassing `BM25SearchTransform`) replaces the single `call_tool` proxy with two:
- `call_read_tool` — read tools only
- `call_write_tool` — write/destructive tools only

Routing keys on the `mutation` alias tag. Because many MCP clients gate permissions by **tool name** (not by the MCP annotation hints), this lets a client allow `call_read_tool` while withholding `call_write_tool` to run the server read-only. `search_tools` still indexes the full catalog.

**Design decisions / scope:**
- **File entities**: creatable only via an `external_url` or an existing `data_file_handle_id` — never from local file content (verified in `synapseclient` that `File.store` only uploads when `path` is set). Metadata/annotation edits and deletion are supported.
- **No wiki-write tool**: the SDK's markdown-store path always writes a temp file to disk, so it is intentionally excluded.
- **Provenance/annotations** are set via `update_entity` (they are entity attributes), not a separate tool.
- **Out of scope**: bulk table row loads (`store_rows`/`upsert_rows`/`update_rows`), FormData and RecordSet creation (all file/CSV-backed), and table row deletion.

# **Testing:**

- **Unit tests**: added per-method service tests (mocked SDK) for all new mutations, plus `tests/test_discovery.py` for the transform (post-transform listing shape; `call_read_tool` rejects mutations and `call_write_tool` rejects reads; both route correctly). Full suite: **378 passed**.
- **BM25 eval**: added a natural-language query fixture per new tool; all per-query top-3 gates pass.
- **Live end-to-end** against the Synapse **dev** backend: create → read → update (rename + annotations + provenance) → delete round-trips for entity, folder, table (+column), external-URL file, ACL set/delete, evaluation, team, and organization/JSON-schema; confirmed no files written to disk and that proxy gating rejects cross-calls. Live testing caught two bugs since fixed: `create_evaluation` was missing required SDK fields, and `delete_acl` falsely reported success on a Project's undeletable root ACL.


src/synapse_mcp/services/entity_service.py
```diff
@@ -412,3 +454,523 @@ async def get_link(
...
+            if name is not UNSET:
+                entity.name = name
+            if parent_id is not UNSET:
+                entity.parent_id = parent_id
+            if description is not UNSET:
+                entity.description = description
+            if annotations is not UNSET:
+                entity.annotations = annotations or {}
+            if provenance is not UNSET:
+                entity.activity = EntityService._build_activity(provenance)
```
BryanFauble, line 808: `update_entity` seems like it is just a small slice of `create_entity` functionality - ie: There are many missing fields that could be updated than is currently updated.

src/synapse_mcp/services/entity_service.py
```diff
@@ -412,3 +454,523 @@ async def get_link(
...
+                        "views, and datasets."
+                    ),
+                    "entity_id": entity_id,
+                }
+            for name in delete_columns or []:
+                entity.delete_column(name=name)
+            for spec in add_columns or []:
+                entity.add_column(EntityService._build_column(spec))
+            stored = await entity.store_async(synapse_client=client)
+            return serialize_model(stored)
```
BryanFauble, line 1089: This would also need some more functionality to update existing bound columns, such as renaming a column, or modifying its order.

src/synapse_mcp/tools.py
```diff
@@ -1641,22 +1604,733 @@ async def search_entities_by_md5(
...
+    description: Optional[str] = None,
+    annotations: Optional[Dict[str, List[Any]]] = None,
+    columns: Optional[List[ColumnSpec]] = None,
+    defining_sql: Optional[str] = None,
+    view_type_mask: Optional[List[ViewScopeType]] = None,
+    scope_ids: Optional[List[str]] = None,
+    target_id: Optional[str] = None,
+    target_version_number: Optional[int] = None,
+    external_url: Optional[str] = None,
+    data_file_handle_id: Optional[str] = None,
```
BryanFauble: There are a whole bunch of arguments here and my 2 concerns:

1. Does an LLM using the tool have a way to enumerate all of the arguments of the tool call?
2. There is no docstring for the tool to get more information on how and when to use each of the arguments? The LLM would also want to return back the docstring for more context.

This might mean we want to expose another tool along side the `call_*` tool call, so something like `describe_tool`.

src/synapse_mcp/services/schema_organization_service.py
```diff
@@ -224,3 +225,147 @@ async def list_json_schema_versions(
...
+            Dict with the created organization metadata.
+        """
+        async with synapse_client(ctx) as client:
+            org = SchemaOrganization(name=organization_name)
+            created = await org.store_async(synapse_client=client)
+            return serialize_model(created)
+
+    @staticmethod
+    @error_boundary(error_context_keys=("organization_name",))
+    async def delete_organization(
```
linglp, line 250: delete an organization should be able to take in either a name or an id

src/synapse_mcp/services/schema_organization_service.py
```diff
@@ -224,3 +225,147 @@ async def list_json_schema_versions(
...
+                "principal_id": principal_id,
+                "access_type": access_type,
+                "updated": True,
+            }
+
+    @staticmethod
+    @error_boundary(
+        error_context_keys=("organization_name", "schema_name")
+    )
+    async def register_json_schema(
```
linglp, line 311: this should be able to take in an organization id or the name

src/synapse_mcp/services/entity_service.py
```diff
@@ -412,3 +454,667 @@ async def get_link(
...
+            target_version_number: Optional pinned version for a link.
+            external_url: For file entities, the external URL to link to.
+            data_file_handle_id: For file/recordset entities, an existing
+                file handle ID to attach.
+
+        Returns:
+            Dict with the created entity's metadata (id, name, etc.).
+        """
+        etype = entity_type.strip().lower().replace("_", "")
+        model = EntityService._build_new_entity(
```
andrewelamb: I think this needs `repository_name` see this script:

```
import asyncio
import os

# Importing synapse_mcp requires an auth env var; the SDK ValueError we
# demonstrate is raised before any network call, so a dummy PAT is fine.
os.environ.setdefault("SYNAPSE_PAT", "dummy")

from synapse_mcp.services.entity_service import EntityService 

# Step 1: the service builds the model. This is the exact call
# create_entity makes (entity_service.py line 506) with the only
# arguments the tool can supply for this type.
model = EntityService._build_new_entity(
    etype="dockerrepository",
    name="my-docker-repo",
    parent_id="syn123456",
    description=None,
    annotations=None,
    columns=None,
    defining_sql=None,
    view_type_mask=None,
    scope_ids=None,
    target_id=None,
    target_version_number=None,
    external_url=None,
    data_file_handle_id=None,
)
print(f"built: {type(model).__name__}")
print(f"repository_name is set: {bool(model.repository_name)}")

# Step 2: the service stores it (entity_service.py line 531).
# The SDK checks repository_name first and raises immediately.
asyncio.run(model.store_async(synapse_client=None))
```

src/synapse_mcp/services/entity_service.py
```diff
@@ -412,3 +454,667 @@ async def get_link(
...
+        return Column(
+            name=spec.get("name"),
+            column_type=col_type,
+            maximum_size=spec.get("maximum_size"),
+            default_value=spec.get("default_value"),
+            enum_values=spec.get("enum_values"),
+        )
+
+    @staticmethod
+    def _resolve_view_mask(masks: List[ViewScopeType]) -> ViewTypeMask:
```
andrewelamb: This returns `None` if the input list is empty.

src/synapse_mcp/services/entity_service.py
```diff
@@ -412,3 +454,667 @@ async def get_link(
...
+                valid = ", ".join(v.name for v in ViewTypeMask)
+                raise ValueError(
+                    f"Unknown view_type_mask '{m}'. Valid: {valid}"
+                )
+            resolved = flag if resolved is None else resolved | flag
+        return resolved
+
+    @staticmethod
+    @error_boundary(error_context_keys=("entity_id",))
+    async def update_entity(
```
andrewelamb: The rule: view_type_mask must not be erased. All lines are in src/synapse_mcp/services/entity_service.py.

What happens with null (correct):
- Line 784 — view_type_mask is in the list of protected fields.
- Lines 787-791 — the check: if the value is None, return the error "cannot be cleared to null." Stop. Nothing is saved.

What happens with [] (the bug):
- Lines 787-791 — the same check looks only for None. An empty list is not None, so it passes.
- Line 830-831 — the value [] is handed to the helper _resolve_view_mask.
- Line 687 — inside the helper: resolved = None.
- Line 688 — the loop runs once per name in the list. The list is empty, so it runs zero times.
- Line 697 — the helper returns resolved, which is still None.
- Line 832 — setattr(entity, "view_type_mask", None). The field is erased.
- Line 837 — store_async saves the entity with the erased field.

So: null is blocked at line 791, but [] sneaks past it and erases the field at line 832 — the exact thing the rule forbids.

See above comment

src/synapse_mcp/services/entity_service.py
```diff
@@ -412,3 +454,667 @@ async def get_link(
...
+
+        Returns:
+            Dict with ``acl_deleted``. A Project's own root ACL cannot be
+            deleted in Synapse, so for a Project ``acl_deleted`` is False
+            and a ``message`` explains how to revoke access instead.
+        """
+        async with synapse_client(ctx) as client:
+            entity = await _resolve_entity(entity_id, client)
+            await entity.delete_permissions_async(synapse_client=client)
+            if type(entity).__name__.lower() == "project":
```
andrewelamb: should this be ` if isinstance(entity, Project):` ?

src/synapse_mcp/services/entity_service.py
```diff
@@ -412,3 +454,667 @@ async def get_link(
...
+                removed so it inherits from its benefactor.
+
+        Returns:
+            Dict with ``acl_deleted``. A Project's own root ACL cannot be
+            deleted in Synapse, so for a Project ``acl_deleted`` is False
+            and a ``message`` explains how to revoke access instead.
+        """
+        async with synapse_client(ctx) as client:
+            entity = await _resolve_entity(entity_id, client)
+            await entity.delete_permissions_async(synapse_client=client)
```
andrewelamb: Should this be moved below to line 955 after the Project check?

src/synapse_mcp/tools.py
```diff
@@ -1641,22 +1610,1080 @@ async def search_entities_by_md5(
...
+    ] = None,
+    data_file_handle_id: Annotated[
+        Optional[str],
+        Field(
+            description="For file/recordset entities, an existing file handle ID, e.g. '9876543'."
+        ),
+    ] = None,
+) -> Dict[str, Any]:
+    """Create a new Synapse entity from metadata."""
+    if parent_id is not None and not validate_synapse_id(parent_id):
```
andrewelamb: Should we do the same check for `target_id` and `scope_ids`?

src/synapse_mcp/tools.py
```diff
@@ -1641,22 +1610,1080 @@ async def search_entities_by_md5(
...
+    if not validate_synapse_id(entity_id):
+        return {
+            "error": f"Invalid Synapse ID: {entity_id}",
+            "error_type": "ValueError",
+        }
+    if (
+        parent_id is not UNSET
+        and parent_id is not None
+        and not validate_synapse_id(parent_id)
+    ):
```
andrewelamb: Should we do the same check for `entity_id`, and `scope_ids`?

src/synapse_mcp/tools.py
```diff
@@ -1641,22 +1610,1080 @@ async def search_entities_by_md5(
...
+            )
+        ),
+    ],
+    ctx: Context,
+    instructions: Annotated[
+        Optional[str], Field(description="Optional free-text instructions.")
+    ] = None,
+) -> Dict[str, Any]:
+    """Create a curation task on a project."""
+    if not validate_synapse_id(project_id):
```
andrewelamb: Should we do the same check for the synapse ids in `task_properties`?

andrewelamb (changes requested):
Hey Bryan! I found a few possible issues to look at.

andrewelamb (approved):
LGTM!

